PR #7, opened 2019-12-13
Settings

apachano:
Implement a system for storing settings values.

Also implement the system into the camera so sensitivity can be adjusted with the p and o keys.

### DIFF
--- a/Engine/Include/Quartz2/Settings.hpp
+++ b/Engine/Include/Quartz2/Settings.hpp
@@ -77,13 +77,13 @@ namespace q2
 		 *
 		 * @return std::string the unique key for the setting
 		 */
-		const std::string getKey();
+		std::string getKey() const;
 		/**
 		 * @brief Gets the value of a setting
 		 *
 		 * @return std::size_t The value of the setting
 		 */
-		const int value();
+		int value() const;
 	};
 
 	class Settings : public Singleton<Settings>
--- a/Engine/Include/Quartz2/Settings.hpp
+++ b/Engine/Include/Quartz2/Settings.hpp
@@ -31,7 +31,7 @@
 #include <Quartz2/Singleton.hpp>
 
 #include <string>
-#include <vector>
+#include <list>
 
 namespace q2 {
 
@@ -71,17 +71,21 @@ namespace q2 {
         /**
          * @brief Gets the value of a setting
          *
-         * @param key The unique name for key ex: core:volume
+         * @return std::string the unique key for the setting
+         */
+        std::string getKey();
+        /**
+         * @brief Gets the value of a setting
+         *
          * @return std::size_t The value of the setting
          */
         int value();
     };
 
     class Settings : public Singleton<Settings> {
-        std::vector<const Setting *> m_setting;
+        std::list<Setting> m_setting;
 
     public:
-        std::size_t add(const Setting *setting);
         /**
          * @brief Adds a new setting
          *
@@ -90,8 +94,8 @@ namespace q2 {
          * @param defaultValue The value the setting will be initially set to
          * @return std::size_t Reuturns the numerical key the setting is stored at
          */
-        std::size_t add(std::string name, std::string key, int defaultValue);
+        Setting *add(std::string name, std::string key, int defaultValue);
 
-        const Setting *getSetting(std::size_t key);
+        Setting *getSetting(std::string key);
     };
 }; // namespace q2
--- a/Engine/Include/Quartz2/Settings.hpp
+++ b/Engine/Include/Quartz2/Settings.hpp
@@ -50,7 +50,7 @@ namespace q2
 		int m_minValue;
 
 	public:
-		Setting(std::string name, std::string key, int defaultValue);
+		Setting(const std::string& name, const std::string& key, int defaultValue);
 		/**
 		 * @brief Sets the value of an already existing setting
 		 *
@@ -99,7 +99,7 @@ namespace q2
 		 * @return std::size_t Reuturns the numerical key the setting is stored
 		 * at
 		 */
-		Setting* add(std::string name, std::string key, int defaultValue);
+		Setting* add(const std::string& name, const std::string& key, int defaultValue);
 
         /**
          * @brief Get the Setting object
@@ -107,6 +107,6 @@ namespace q2
          * @param key Unique Name for key ex: core:volume
          * @return Setting* A pointer to the setting object
          */
-		Setting* getSetting(std::string key);
+		Setting* getSetting(const std::string& key);
 	};
 }; // namespace q2
--- a/Engine/Include/Quartz2/Settings.hpp
+++ b/Engine/Include/Quartz2/Settings.hpp
@@ -76,13 +76,13 @@ namespace q2
 		 *
 		 * @return std::string the unique key for the setting
 		 */
-		std::string getKey();
+		const std::string getKey();
 		/**
 		 * @brief Gets the value of a setting
 		 *
 		 * @return std::size_t The value of the setting
 		 */
-		int value();
+		const int value();
 	};
 
 	class Settings : public Singleton<Settings>
@@ -107,6 +107,6 @@ namespace q2
 		 * @param key Unique Name for key ex: core:volume
 		 * @return Setting* A pointer to the setting object
 		 */
-		const Setting* getSetting(const std::string& key);
+		Setting* getSetting(const std::string& key);
 	};
 }; // namespace q2
--- a/Engine/Include/Quartz2/Settings.hpp
+++ b/Engine/Include/Quartz2/Settings.hpp
@@ -50,7 +50,7 @@ namespace q2
 		int m_minValue;
 
 	public:
-		Setting(const std::string& name, const std::string& key, int defaultValue);
+		Setting(std::string name, std::string key, int defaultValue);
 		/**
 		 * @brief Sets the value of an already existing setting
 		 *
--- a/Engine/Include/Quartz2/Settings.hpp
+++ b/Engine/Include/Quartz2/Settings.hpp
@@ -33,63 +33,63 @@
 #include <string>
 #include <vector>
 
-namespace q2{
+namespace q2 {
 
-    struct Setting
-    {
-        /// @brief Human readable name for setting
-        std::string name;
-        /// @brief Unique name for key ex: core:volume
-        std::string key;
-        /// @brief Value of setting
-        std::size_t value;
+    struct Setting {
+    /// @brief Human readable name for setting
+    std::string name;
+    /// @brief Unique name for key ex: core:volume
+    std::string key;
+    /// @brief Value of setting
+    std::size_t value;
     };
 
-    class Settings : public Singleton<Settings>
-    {
-        std::vector<Setting> m_setting;
+    class Settings : public Singleton<Settings> {
+    std::vector<Setting> m_setting;
 
     public:
-        /**
-         * @brief Adds a new setting
-         * 
-         * @param name Human readable name for setting
-         * @param key Unique Name for key ex: core:volume
-         * @param defaultValue The value the setting will be initially set to
-         * @return std::size_t Reuturns the numerical key the setting is stored at
-         */
-        std::size_t add(std::string name, std::string key, std::size_t defaultValue);
-        /**
-         * @brief Sets the value of an already existing setting
-         * 
-         * @param value The value to be set
-         * @param key The unique name for key ex: core:volume
-         * @return true if the setting was set
-         * @return false if the setting does not exist
-         */
-        bool set(std::size_t key, std::size_t value);
-        /**
-         * @brief Sets the value of an already existing setting
-         * 
-         * @param value The value to be set
-         * @param key The numerical key the setting is stored at
-         * @return true if the setting was set
-         * @return false if the setting does not exist
-         */
-        bool set(std::string key, std::size_t value);
-        /**
-         * @brief Gets the value of a setting
-         * 
-         * @param key The unique name for key ex: core:volume
-         * @return std::size_t The value of the setting
-         */
-        std::size_t value(std::string key);
-        /**
-         * @brief Gets the value of a setting
-         * 
-         * @param key The numerical key the setting is stored at
-         * @return std::size_t The value of the setting
-         */
-        std::size_t value(std::size_t key);
-    };
+    /**
+     * @brief Adds a new setting
+     *
+     * @param name Human readable name for setting
+     * @param key Unique Name for key ex: core:volume
+     * @param defaultValue The value the setting will be initially set to
+     * @return std::size_t Reuturns the numerical key the setting is stored at
+     */
+    std::size_t add(std::string name, std::string key, std::size_t defaultValue);
+    /**
+     * @brief Sets the value of an already existing setting
+     *
+     * @param value The value to be set
+     * @param key The unique name for key ex: core:volume
+     * @return true if the setting was set
+     * @return false if the setting does not exist
+     */
+    bool set(std::size_t key, std::size_t value);
+    /**
+     * @brief Sets the value of an already existing setting
+     *
+     * @param value The value to be set
+     * @param key The numerical key the setting is stored at
+     * @return true if the setting was set
+     * @return false if the setting does not exist
+     */
+    bool set(std::string key, std::size_t value);
+    /**
+     * @brief Gets the value of a setting
+     *
+     * @param key The unique name for key ex: core:volume
+     * @return std::size_t The value of the setting
+     */
+    std::size_t value(std::string key);
+    /**
+     * @brief Gets the value of a setting
+     *
+     * @param key The numerical key the setting is stored at
+     * @return std::size_t The value of the setting
+     */
+    std::size_t value(std::size_t key);
+
+    Setting *getSetting(std::size_t key);
 };
+}; // namespace q2
--- a/Engine/Include/Quartz2/Settings.hpp
+++ b/Engine/Include/Quartz2/Settings.hpp
@@ -87,7 +87,7 @@ namespace q2
 
 	class Settings : public Singleton<Settings>
 	{
-		std::list<Setting> m_setting;
+		std::list<Setting> m_settings;
 
 	public:
 		/**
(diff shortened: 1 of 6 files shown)


apachano:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/GentenStudios/Phoenix/pullRequest?prid=4673308&bid=15638527)

apachano:
Moving back to WIP, storing the setting outside the system is causing issues (and opens up to invalidating pointers) Need to revise.

Engine/Source/Settings.cpp
```diff
@@ -0,0 +1,74 @@
...
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+#include <Quartz2/Settings.hpp>
+#include <climits>
+
+using namespace q2;
+
+Setting::Setting(const std::string& name, const std::string& key, int defaultValue)
+    : m_name(std::move(name)), m_key(std::move(key)), m_value(defaultValue), m_maxValue(SHRT_MAX),
```
beeperdeeper089: Don't move a const reference. Either move a raw ``std::string`` or just make it equal the ``const std::string&``

Engine/Include/Quartz2/Settings.hpp
```diff
@@ -0,0 +1,112 @@
...
+		 */
+		Setting* add(const std::string& name, const std::string& key, int defaultValue);
+
+        /**
+         * @brief Get the Setting object
+         * 
+         * @param key Unique Name for key ex: core:volume
+         * @return Setting* A pointer to the setting object
+         */
+		Setting* getSetting(const std::string& key);
```
DontBelieveMe: Indentation here seems to be off (1 tab too short I think?)

Engine/Source/Settings.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+bool Setting::set(int value)
+{
+	if (m_maxValue >= value && value >= m_minValue)
+	{
+		m_value = value;
+		return true;
+	}
+	return false;
+};
```
DontBelieveMe: Unnecessary semicolon at the end of a function body.

Engine/Source/Settings.cpp
```diff
@@ -0,0 +1,74 @@
...
+using namespace q2;
+
+Setting::Setting(std::string name, std::string key, int defaultValue)
+    : m_name(std::move(name)), m_key(std::move(key)), m_value(defaultValue), m_maxValue(SHRT_MAX),
+      m_minValue(SHRT_MIN)
+{}
+
+bool Setting::set(int value)
+{
+	if (m_maxValue >= value && value >= m_minValue)
```
DontBelieveMe: I would write this as
```cpp
if (value >= m_minValue && value <= m_maxValue)
```
I think it is clearer that the value must sit within the range (min, max), as I had to think twice when looking at that if statement.

Engine/Source/Settings.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+Setting* Settings::add(const std::string& name, const std::string& key, int defaultValue)
+{
+	m_setting.push_back(Setting(name, key, defaultValue));
+	return &m_setting.back();
+};
+
+Setting* Settings::getSetting(const std::string& key)
+{
+	for (std::list<Setting>::iterator it = m_setting.begin();
```
DontBelieveMe: You can just use a for-range loop here, avoid touching standard library iterators
```cpp
for (const Setting& setting : m_setting) { ... }
```

Engine/Source/Settings.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+std::string Setting::getKey() { return m_key; };
+
+int Setting::value() { return m_value; };
+
+Setting* Settings::add(const std::string& name, const std::string& key, int defaultValue)
+{
+	m_setting.push_back(Setting(name, key, defaultValue));
+	return &m_setting.back();
+};
```
DontBelieveMe: Unnecessary semicolon

Engine/Source/Settings.cpp
```diff
@@ -0,0 +1,74 @@
...
+	for (std::list<Setting>::iterator it = m_setting.begin();
+	     it != m_setting.end(); ++it)
+	{
+		if (it->getKey() == key)
+		{
+			return &(*it);
+		}
+	}
+	return nullptr;
+};
```
DontBelieveMe: Unneeded semicolon

Engine/Include/Quartz2/Settings.hpp
```diff
@@ -0,0 +1,112 @@
...
+		 * @brief Gets the value of a setting
+		 *
+		 * @return std::size_t The value of the setting
+		 */
+		int value();
+	};
+
+	class Settings : public Singleton<Settings>
+	{
+		std::list<Setting> m_setting;
```
DontBelieveMe: Since `m_setting` is a list I would call it `m_settings` to indicate that it represents multiple settings.

Engine/Include/Quartz2/Settings.hpp
apachano: Weird, somehow those ended up being spaces instead of tabs

Engine/Source/Settings.cpp
apachano: @DontBelieveMe Would this make any sense?
`if (m_minValue <= value && value <= m_maxValue)`
I can't remember what language allows this but I originally wanted
`if (m_minValue <= value <= m_maxValue)`

DontBelieveMe: That expression makes sense from a pure mathematical point of view (where such is nicely represented as `x < y < z` or something like that to represent range.
I don't know which language supports that (maybe python? - it likes nice syntactical things like that).

 `m_minValue <= value && value <= m_maxValue` would work, but I don't think it makes much sense to write it that way from a programming point of view - where it's broken down into, `value` is greater than `m_minValue` and `value` is smaller than `m_maxValue`.

So I'd write it as `value >= m_minValue && value <= m_maxValue` still, I have to think twice when reading `m_minValue <= value && value <= m_maxValue` in order to understand what it means.

apachano: @DontBelieveMe Makes sense, updated

apachano:
@DontBelieveMe @beeperdeeper089 This should be all set for re-review

Engine/Source/Settings.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+Setting* Settings::add(const std::string& name, const std::string& key, int defaultValue)
+{
+	m_settings.push_back(Setting(name, key, defaultValue));
+	return &m_settings.back();
+}
+
+const Setting* Settings::getSetting(const std::string& key)
+{
+	for (Setting& setting : m_settings)
```
DontBelieveMe: Since nothing in setting appears to be getting modified, this could be a const reference.
```suggestion
	for (const Setting& setting : m_settings)
```

Engine/Include/Quartz2/Settings.hpp
```diff
@@ -0,0 +1,112 @@
...
+		 *
+		 * @param value The minimum value the setting can be
+		 */
+		void setMin(int value);
+		/**
+		 * @brief Gets the value of a setting
+		 *
+		 * @return std::string the unique key for the setting
+		 */
+		std::string getKey();
```
DontBelieveMe: This function can be marked as `const` (since it's a getter & does not modify anything)
```suggestion
		std::string getKey() const;
```

Engine/Include/Quartz2/Settings.hpp
```diff
@@ -0,0 +1,112 @@
...
+		/**
+		 * @brief Gets the value of a setting
+		 *
+		 * @return std::size_t The value of the setting
+		 */
+		int value();
```
DontBelieveMe: This function also could be `const`.
```suggestion
		int value() const;
```

Engine/Source/Settings.cpp
```diff
@@ -0,0 +1,73 @@
...
+	return false;
+}
+
+void Setting::setMax(int value) { m_maxValue = value; }
+
+void Setting::setMin(int value) { m_minValue = value; }
+
+std::string Setting::getKey() { return m_key; }
+
+int Setting::value() { return m_value; }
```
DontBelieveMe: This would have to be updated if `Setting::value` is made const
```suggestion
int Setting::value() const { return m_value; }
```

Engine/Source/Settings.cpp
```diff
@@ -0,0 +1,73 @@
...
+		return true;
+	}
+	return false;
+}
+
+void Setting::setMax(int value) { m_maxValue = value; }
+
+void Setting::setMin(int value) { m_minValue = value; }
+
+std::string Setting::getKey() { return m_key; }
```
DontBelieveMe: This method implementation would also have to be updated if `Setting::getKey()` is made const
```suggestion
std::string Setting::getKey() const { return m_key; }
```

Engine/Include/Quartz2/Settings.hpp
```diff
@@ -0,0 +1,112 @@
...
+		int m_minValue;
+
+	public:
+		Setting(std::string name, std::string key, int defaultValue);
+		/**
+		 * @brief Sets the value of an already existing setting
+		 *
+		 * @param value The value to be set
+		 * @return true if the setting was set
+		 * @return false if the value was not within the settings max/min
```
DontBelieveMe: Is this system of specifying multiple `@return`s actually valid doxygen? If it is, that's awesome and nothing to change here, but I always thought you could only have one - so it would have to be
```
@return true if the setting is set, false if the value was not within the settings max/min.
```

DontBelieveMe:
Again, really sorry for leaving so many comments on a rereview, I should really pick them up first time round, but I just spot things or have questions etc when doing another review.
So yeah, some more comments 😟 , sorry! 😄

Engine/Source/Settings.cpp
DontBelieveMe: 👍 Looks good

Engine/Source/Settings.cpp
```diff
@@ -0,0 +1,73 @@
...
+const Setting* Settings::getSetting(const std::string& key)
+{
+	for (Setting& setting : m_settings)
+	{
+		if (setting.getKey() == key)
+		{
+			return &setting;
+		}
+	}
+	return nullptr;
```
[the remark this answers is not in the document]
apachano: @DontBelieveMe std::find (and friends) won't work because they compare the settings objects, we would need to know what value the setting is at so we can set the comparison value (which defeats the purpose). However, an unordered_map would solve this anyways.

Engine/Include/Quartz2/Settings.hpp
```diff
@@ -0,0 +1,112 @@
...
+		/// @brief Human readable name for setting
+		std::string m_name;
+		/// @brief Unique name for key ex: core:volume
+		std::string m_key;
+		/// @brief Value of setting
+		int m_value;
+		/// @brief Maximim value that the setting can be set to
+		int m_maxValue;
+		/// @brief Minimum value that the setting cab be set to
+		int m_minValue;
```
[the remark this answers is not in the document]
apachano: We would do a map that any system requiring the setting can access, alternatively, some sort of map built into the setting could be used. I would love a simpler alternative but I couldn't think of an easy one nor could I come up with a situation where a setting couldn't be mapped to an integer. (Input mapping will be its own system)

Engine/Include/Quartz2/Settings.hpp
apachano: @DontBelieveMe TBH my IDE did that automatically so I assume yes (also it pulls that cleanly in the helper when using the function).

Engine/Include/Quartz2/Settings.hpp
```diff
@@ -0,0 +1,112 @@
...
+		 * @brief Gets the value of a setting
+		 *
+		 * @return std::size_t The value of the setting
+		 */
+		int value();
+	};
+
+	class Settings : public Singleton<Settings>
+	{
+		std::list<Setting> m_settings;
```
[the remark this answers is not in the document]
apachano: @DontBelieveMe You're right on point, I think a map would make sense both for speed and simplicity. Do we want to do the same with the Block registry as well since it's in the same boat?

Engine/Source/Settings.cpp
apachano: It invalidates the call to getKey(), we get this error on compiling:
`the object has type qualifiers that are not compatible with the member function "q2::Setting::getKey" -- object type is: const q2::Setting`

That being said, if we want to refactor this to use a map, we are no longer using this search method.

apachano:
@DontBelieveMe No problem, I like the in-depth reviews. We are getting solid code together and my ability to code has improved greatly. That being said some of the stuff you find could be automated, I would really like to see if we can do that to speed up the review process. (If a robot caught some of these, I could have addressed them day 1).

Engine/Source/Settings.cpp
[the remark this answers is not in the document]
DontBelieveMe: yeah that's what std::find does but std::find_if has a function parameter which is used as a comparator - eg in this case you'd have a lambda that compares the key of the setting with the input key or something like that

Engine/Source/Settings.cpp
DontBelieveMe: Yeah that's because `getKey` is non const, if the method was made const as per a previous comment then you wouldn't get this error

Engine/Include/Quartz2/Settings.hpp
DontBelieveMe: Yeah you're right, it's valid
```
Multiple adjacent \return commands will be joined into a single paragraph. 
```
From the doxygen documentation.

Engine/Source/Settings.cpp
apachano: @DontBelieveMe I already fixed the getKey() but it's still not happy :(. Regardless I think I am going to attempt the map this morning.

Engine/Include/Quartz2/Settings.hpp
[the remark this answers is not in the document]
apachano: @DontBelieveMe Updated to a map. Was easier than I thought except it throws errors if a blank constructor doesn't exist. Not sure how to get around that.

apachano:
@DontBelieveMe FWIW the remaining open comments have been addressed as I swapped the list out with a map. The code is actually significantly simpler now :D

Would you mind doing a re-review so we can see if this can get closed out this week?

Engine/Source/Settings.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+Setting* Settings::add(const std::string& name, const std::string& key, int defaultValue)
+{
+	m_settings[key] = Setting(name, key, defaultValue);
+	return &m_settings[key];
+}
+
+Setting* Settings::getSetting(const std::string& key)
+{
+	return &m_settings[key];
```
DontBelieveMe: So `operator[]` has a strange thing where if `key` does not exist in the map then it add's a default constructed value (in this case a `Setting`).
The way to get around this is to either use `.at` (which I believe throws an exception if the key does not exist, or manually check ourselves.
In this case I'd suggest manually checking the key is in the map and if so, returning `nullptr`.

Something like this
```cpp
if (m_settings.find(key) != m_settings.end())
    return &m_settings[key];
else
    return nullptr;
```

Engine/Include/Quartz2/Settings.hpp
```diff
@@ -0,0 +1,113 @@
...
+		 *
+		 * @param value The minimum value the setting can be
+		 */
+		void setMin(int value);
+		/**
+		 * @brief Gets the value of a setting
+		 *
+		 * @return std::string the unique key for the setting
+		 */
+		const std::string getKey();
```
DontBelieveMe: Ah, by making this a const function, I don't mean return a const value, but make the function itself const.

```cpp
std::string getKey() const; // <-- Note the const comes after the ()
```

This just means that we can guarantee that what is in this function does not modify the state of the object, and as such it can be called on const objects.
e.g.

```cpp
const Setting setting(...);

// We can use getKey here if it is declared as const (std::string getKey() const)
// If the const is omitted from the end of the function signature then this would cause an error.
std::string key = setting.getKey(); 
```
Personally I wouldn't make the return value const, just to leave it up to the caller, but that's a different discussion.

Engine/Include/Quartz2/Settings.hpp
```diff
@@ -0,0 +1,113 @@
...
+		/**
+		 * @brief Gets the value of a setting
+		 *
+		 * @return std::size_t The value of the setting
+		 */
+		const int value();
```
DontBelieveMe: I would also make this a `const` function (by appending `const` after the `()` as talked about in a previous comment). Just as it's a getter and so does not modify the object state in any way.

Engine/Source/Settings.cpp
```diff
@@ -0,0 +1,68 @@
...
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+#include <Quartz2/Settings.hpp>
+#include <climits>
+
+using namespace q2;
+
+Setting::Setting(){};
```
DontBelieveMe: I feel a bit like a stuck record, but until we get this automated - unnecessary semicolon after the `{}` - sorry 😟 😄

Engine/Include/Quartz2/Settings.hpp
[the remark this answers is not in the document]
DontBelieveMe: Yeah I guess, anyhow this is not a big blocker & can be expanded on later when it's merged.

DontBelieveMe:
Right, a few more things but after those changes, it looks good & can be merged 👍

beeperdeeper089:
this is a test for the discord webhook

beeperdeeper089:
this is another test for the discord webhook

apachano:
@DontBelieveMe Fixed and it works! Could I get that approving review now?

DontBelieveMe:
Lgtm as far as I'm concerned it can be merged, unless anyone else has any comments

apachano:
@beeperdeeper089 could you take a final look and complete your re-review so we can close this out?

PR #8, opened 2019-12-14
Setup CI

apachano:


.appveyor.yml
```diff
@@ -0,0 +1,7 @@
+image: Visual Studio 2017
+matrix:
+    fast_finish: true
+
+build_script:
+    - cmake -H. -Bbuild
+    - cmake --build build
```
DontBelieveMe: Add `--target` here so we only build what we need to build to game.
```suggestion
    - cmake --build build --target Sandbox
```

.travis.yml
```diff
@@ -0,0 +1,14 @@
...
+    - linux
+    - osx
+
+compiler:
+    - gcc
+    - clang
+
+script:
+    - cmake -H. -Bbuild
+    - cmake --build build
```
DontBelieveMe: Same here.
```suggestion
    - cmake --build build --target Sandbox
```

DontBelieveMe:
@apachano Can this be closed now? I'll make another separate PR for Travis when I've set that up

PR #20, opened 2019-12-20
Switch to using texture arrays instead of building texture atlas at runtime.

DontBelieveMe:
We currently build a big texture atlas of all textures forming a single OpenGL texture which is accessed by having different UVs for each texture. This worked but was a bit slow and resulted in some visual artefacts when looking at chunks at a certain angle (I suspect due to the textures leaking into each other in the atlas).
This patch moves from that system to use OpenGL texture arrays to store block textures.


Engine/Include/Quartz2/ChunkRenderer.hpp
```diff
@@ -72,10 +72,12 @@ namespace q2
 		void addMesh(const std::shared_ptr<Mesh>& mesh);
 
 		void setTexture(unsigned char* data, std::size_t w, std::size_t h);
+		void setTextureArray(GLuint textureArray);
```
beeperdeeper089: Maybe don't use GLuint here, but just a uint and convert in the definition? (just for standardisation)

Engine/Source/BlocksTextureAtlas.cpp
```diff
@@ -34,133 +34,125 @@
...
+	// Create the texture array, but don't fill it in with
+	// anything - that'll be done by glTexSubImage3D calls
+	// for each texture later.
+	glTexImage3D(
+		GL_TEXTURE_2D_ARRAY,             // Texture Target.
+		0,                               // LoD number - 0 is the base image level.
+		GL_RGBA8,                        // Internal format - Number of color components.
+		(GLsizei) m_spriteWidth,         // Width of the textures in the array (pixels).
+		(GLsizei) m_spriteHeight,        // Height of the textures in the array (pixels).
+		(GLsizei) m_textureIDMap.size(), // Number of textures in the array (depth/number of layers).
```
beeperdeeper089: We can't guarantee this will always work, since OpenGL 3.3 only says there should be a minimum of 256, and the chances of us having more than that is very high, so if we implement a check here, that would probably be beneficial to finding out why something isn;'t working and this is the case.

Engine/Include/Quartz2/ChunkRenderer.hpp
DontBelieveMe: Yeah this makes sense - In fact I think it makes sense to define `GLObjectHandle` as `GLuint` or something (To make it documenting that this is a handle to an OpenGL object) - If that sounds OK?

Engine/Source/BlocksTextureAtlas.cpp
DontBelieveMe: Yeah need to do this - What we should probably do is just create more texture array objects if we overflow the max number of layers in one.
That approach would be a fairly major change to the whole process, so for now I'll just error (however we end up doing that) if we have more textures, if that sounds good?

beeperdeeper089: Yeah, just erroring for now should be fine.

Engine/Include/Quartz2/ChunkRenderer.hpp
beeperdeeper089: SOunds good to me.

DontBelieveMe:
Err I don't know why I closed this... 🤦‍♂

Engine/Include/Quartz2/ChunkRenderer.hpp
DontBelieveMe: Added `OpenGL32::Object` type alias for this (aliases to `GLuint`). I've updated it for the texture arrays stuff, there is more that could use it, but that's a different PR.

Engine/Source/BlocksTextureAtlas.cpp
DontBelieveMe: Done this - until we decide on an approach for errors, do a hard crash in debug mode & just fail in release mode.

Engine/Include/Quartz2/BlocksTextureAtlas.hpp
```diff
@@ -29,54 +29,106 @@
...
+		SpriteID    getSpriteIDFromFilepath(const char* filepath);
 
-		RectAABB getSpriteFromID(SpriteID spriteId) const;
+		/**
+		 * @brief Get the OpenGL ID of the texture array. Valid only after
+		 *        patch() has been called.
+		 *
+		 * @return OpenGL ID of the texture array (generated by patch())
+		 */
+		OpenGL32::Object getTextureArrayID() const { return m_textureArray; }
```
beeperdeeper089: Do we need the 32? It would probably look/feel a bit cleaner without it IMO; I'm fine with it being there, but just a question as to your thoughts behind it.

Engine/Source/BlocksTextureAtlas.cpp
```diff
@@ -34,133 +34,127 @@
...
-		equalsTest) == m_textureIDMap.end())
-		return BlockTextureAtlas::INVALID_SPRITE;
-
-	return m_textureIDMap.at(filepath);
+	m_spriteHeight = h;
 }
 
-RectAABB BlockTextureAtlas::getSpriteFromID(
-	BlockTextureAtlas::SpriteID spriteID) const
+void BlockTextureAtlas::addTextureFile(const char* texturefilepath)
```
beeperdeeper089: Should we add an overload for an ``std::string`` that will just do a ``.c_str()`` and call this function? Might be redundant, but it might be useful when integrating it with the Lua system (idk how strings will work there, just food for thought).

Engine/Source/BlocksTextureAtlas.cpp
```diff
@@ -34,133 +34,127 @@
...
+	glTexParameteri(GL_TEXTURE_2D_ARRAY, GL_TEXTURE_MAG_FILTER, GL_NEAREST);
+	glTexParameteri(GL_TEXTURE_2D_ARRAY, GL_TEXTURE_WRAP_S, GL_CLAMP_TO_EDGE);
+	glTexParameteri(GL_TEXTURE_2D_ARRAY, GL_TEXTURE_WRAP_T, GL_CLAMP_TO_EDGE);
 
-BlockTextureAtlas::~BlockTextureAtlas() { delete[] m_patchedTextureData; }
\ No newline at end of file
+	// Unbind the texture now we're done with it (for generation).
+	// Not strictly necessary, but shouldn't have to much of a performance impact
+	// here.
+	glBindTexture(GL_TEXTURE_2D_ARRAY, 0);
```
beeperdeeper089: Do we need to generate mipmaps? (just since it's not there, idk if we should.)

Engine/Source/Chunk.cpp
```diff
@@ -40,98 +41,118 @@ Chunk::Chunk()
...
+static Vec3 getTexCoords(const Vec2& uvs, BlockTextureAtlas::SpriteID id)
+{
+	return { uvs.x, uvs.y, id };
 }
 
 static void addBlockTopFace(Vec3 pos, Mesh* mesh, BlockType* type)
 {
-	const RectAABB uvs = BlockRegistry::get()->getAtlas()->getSpriteFromID(type->textures.top);
+	const RectAABB uvs = getRectUVs();
+	const BlockTextureAtlas::SpriteID spriteId = type->textures.top;
```
beeperdeeper089: Variable should be ``spriteID`` to match with current coding standards.

Engine/Source/ChunkRenderer.cpp
```diff
@@ -62,26 +64,17 @@ void ChunkRenderer::addMesh(const std::shared_ptr<Mesh>& mesh)
 	glBufferData(GL_ARRAY_BUFFER, vertices.size() * sizeof(Vertex), vertices.data(), GL_STATIC_DRAW);
 
 	glVertexAttribPointer(0, 3, GL_FLOAT, GL_FALSE, sizeof(Vertex), (void*)offsetof(Vertex, position));
-	glVertexAttribPointer(1, 2, GL_FLOAT, GL_FALSE, sizeof(Vertex), (void*)offsetof(Vertex, texCoord));
+	glVertexAttribPointer(1, 3, GL_FLOAT, GL_FALSE, sizeof(Vertex), (void*)offsetof(Vertex, texCoord));
```
beeperdeeper089: Use the explicit cast (probably ``reinterpret_cast`` in this scenario).

Sandbox/Assets/TerrainShader.frag
```diff
@@ -1,11 +1,12 @@
 #version 330 core
```
beeperdeeper089: We should upgrade shader versions and start using explicit positioning through the ``layout`` attribute in here, or set that through shader initialisation for internal tracking in the shader compilation code.

Sandbox/Assets/TerrainShader.frag
```diff
@@ -1,11 +1,12 @@
...
 
 in vec3 pass_Position;
-in vec2 pass_TexCoord;
+in vec3 pass_TexCoord;
 out vec4 out_FragColor;
 uniform sampler2D u_Sampler;
```
beeperdeeper089: Dump this if we don't need it.

beeperdeeper089:
Solved through introduction of new texture registry and chunk renderer.

PR #30, opened 2020-01-04
Add OSX image to travis

apachano:


apachano:
@beeperdeeper089 I am okay with this as a fix. Turns out Apple pulled an Apple and built their own proprietary clang that is slightly out of date from llvm (despite being slightly more stable) but <optional> is available in the latest version.

Issue #19, opened 2019-12-20
In game terminal


apachano:
## Story
In order to execute commands, the game needs a built-in terminal that accepts input and outputs.

## MVP
What are the minimum features needed for this story?
- [ ] Pressing the enter key brings up edit field and output box
- [x] Output from a stream is displayed above edit field similar to a terminal
- [x] Hitting return sends the contents of the edit field to a stream
- [ ] When in edit mode scrolling scrolls back the history of the output display

## Stretch
What are some non-required features for this story?
- [x] The output still displays but with no background and less scrollback when the edit box is not focused
- [ ] Text highlighting is possible in the output window
- [ ] Terminal history recall

M3TIOR:
I'm working on this rn & have been for the last few days. Assign me to the job pleeez! :blue_heart:

tobyplowy:
@M3TIOR It has been done.
Glad to have you on this and good luck :)

M3TIOR:
@apachano what kind of highlighting were you thinking? As in: automated highlighting, or user directed / selection type highlighting?

apachano:
@M3TIOR I was thinking ANSI escape code compatibility is all, from there the program can handle the rest. That will also make it compatible with the terminal when executing commands via the server.

M3TIOR:
Okay thank god, that's what I thought you meant and what I started adding scaffolding for, but just wanted to make sure. Thx! :+1: :smiley:

M3TIOR:
Also what do you mean by "edit mode"?

apachano:
"edit mode" is when the cursor is focused on the input field. In terms of the MVP this is when the terminal is visible.

M3TIOR:
Okay, good, we're on the same page then. From the way you initially said it, it sounded like maybe you were talking about terminal history recall, but that's not really an extremely useful feature atm. Just wanted to make sure.

apachano:
No, definitely not critical right now. Although I added it as a stretch item (mostly to remind myself to add it as a low priority issue when this closes out).

Issue #4, opened 2019-12-11
ImGUI Lua Implementation


apachano:
## Story
As a developer, I want to be able to add custom GUI elements to my game. This could be in the form of a quick command menu, a settings menu, or even the crafting system. 
## MVP
I should be able to create a window, with a text box, a button, a checkbox, a radial, and a dropdown.

Issue #3, opened 2019-12-11
Chunk Meshing


apachano:
## Story
Chunks that are next to each other should mesh together so you cant tell they are separate chunks.

PR #38, opened 2020-01-05
Feat ingame terminal

M3TIOR:
# The In Game Terminal Is Here!
## From the checklist issue #19 this resolves the following:
 - [x] Pressing the enter key brings up edit field and output box
 - [x] Output from a stream is displayed above edit field similar to a terminal
 - [x] Hitting return sends the contents of the edit field to a stream
 - [x] When in edit mode scrolling scrolls back the history of the output display
 - [x] The output still displays but with no background and less scrollback when edit box is not focused

## Caveats
 - The edit field and output box must have the mouse hovering over them for scrolling to work for ImGui internal implementation reasons. This will be the case until we can find a workaround.
-  Using the enter key functionality to set focus on the chat terminal is disabled in comments right now because it's not capable of overlooking other terminal widgets once active. A more detailed explanation can be found in my comments.
- For some reason when entering a huge string sometimes the terminal doesn't know how to handle it. There could definitiely be some improvements made to my output cache flush method.


apachano:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
- Added 5
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/GentenStudios/Phoenix/pullRequest?prid=4764724&bid=15884389)

M3TIOR:
I'd recommend meld for merge conflicts. Their GUI is fucking amazing.

apachano:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/GentenStudios/Phoenix/pullRequest?prid=4764724&bid=15884389)

M3TIOR:
Finally, butt == kicked Codacy you nasty devil. Go jump off a bridge and swim to shore covered in seaweed, half drowned like the nasty swamp urchin you are.

Harumph :triumph:

M3TIOR:
Why do I spend so much of my time on git messages when nobody reads them? Oh well.

apachano:
@M3TIOR Codacy down, still need to answer to Appveyor and you have yet to awaken Le Travis. Merge develop into your code and we shall see.

M3TIOR:
:martial_arts_uniform: Hiyah! Take that! :stuck_out_tongue_closed_eyes: Eyes closed in case I miss again and wind up on my ass looking like a chump to these behemoths.

M3TIOR:
Why you hate me appveyor?

M3TIOR:
Xcode really really hates me. I'm recompiling everything now that I've pulled the upstream changes. I probably should have done that before though.

apachano:
@M3TIOR Still pre-review but caveat 3 and 4 also sound significant. 
- How bad is the overlap? This may not be a blocker but would definitely be a serious bug.
- If pressing enter does not focus, how do you focus/ free the cursor? 
For 5, any idea how long huge is?

beeperdeeper089:
I'm assuming we could overcome No.4 with the Layer Stack system I proposed over Discord, by forcing focus because only specific widgets would be active at any given time.

M3TIOR:
I agree with @beeperdeeper089, if we have a way of determining what widgets must be active at all times, it would keep the focus method pointing at the right widget and thus we shouldn't have that problem. As for the #3 overlap, It's about 30 characters, and only happens when wrapping a contiguous stream of characters because ImGui is programmed to break to newlines when there are spaces in a string that's long enough. The head portion of the extremely wrong word / stream is what gets overlapped so if users aren't typing anything too redicuous like `AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA` or rage spamming like myself `shoviuahbvopuiewb[o iaw[ pwian[o ugawh[oekvjbpawougyv [wiegvp9[euvg` everything should be fine. With #4 I've had a hard time replicating the crash results because they're fairly inconsistent on my end. It seems to be as a result of me not handling an `out_of_bounds exception` when trying to transfer the string contents from one output stream to then next. Like I said, probs a problem with my algorithm that is easy to fix. I went ahead and pushed this anyway though because unless users are putting in strings like the ones above but way waaaaaay longer, there aren't any problems thus far.

beeperdeeper089:
Because I want to actually run this code so it makes sense (for review), are there any special test cases/features you want to me to storm test to ensure reliability? Otherwise I can get straight to reviewing just code standards/quality if you're confident it should work for the most part.

M3TIOR:
I'm confident it should work for the most part. As of right now I forgot to mention that I haven't set up any overrides for the onCommand event so outputting to the terminal is in raw append mode for now. Ie. Everything you type will be appended to the really long string at the end without newlines and I don't have a way to put newlines in atm. Newlines vs no newlines would be a good test to see if that text rendering bug is not just something that happens on my machine. You should also be aware that the last rendered line will always have the text "This is an ImGui Bug" as a test for the text render nonesense. I should probably have written a test method to output a large glob of text to the terminal's cout streams just to make sure my stream emptying method works fine.

M3TIOR:
Whoops should have been a damn refactor lable instead of a fix because I didn't actually change the code definitions I just moved the inlines into the headers where they're supposed to be. Argh. :anger:

apachano:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/GentenStudios/Phoenix/pullRequest?prid=4764724&bid=15884389)

M3TIOR:
Whenever you're ready to review @beeperdeeper089, it's ready to merge.

apachano:
Yea, it's there although we need to finish those.

Other notes, namespaces should start with a lower case (although we need to sit down and redo the namespacing at some point soon.

And should we be editing code inside a third party? I think the longterm goal is to do a submodule for ImGUI but even if we don't, custom code could be missed if we update.

M3TIOR:
I figured there'd be a doxy issue,  I'll get on that when I feel better I'm super exhausted rn. I was planning on doing that before I pushed initially but figured it could waiit because we've got a hidden deadline for the voxel code, which I wanted to stsrt working on if Bladk is going to be eternally absent. 

As for the ImGui edits, that'll probably be a PR from my own ImGui fork. Updating it should be easy once the PR is merged. That's more of a long term goal though.

M3TIOR:
Since we're using camelCase and the ImGui namespace was already capitalized, I assumed that the trend should carry over into our code. Besides that, while I know ImGui uses snake_case for it's code style so there's a conflict of interest there, I feel it important to keep the `ImGui` name case consistent. It just makes more sense in the english language because it's the name of an API / project. 

If you feel those arguments are insubstantial I'll change the namespace before the pull, but please consider that the inconsistency will also make it harder and more frustrating for me to use my own system :sweat_smile: .

apachano:
I'll defer to @beeperdeeper089 for the style argument, IMO every third party will have its own style, if we try to match all of them we will end up with a mix of styles all over the place. For code written in the project I would think that would want to match the style of the rest of the code.

Doxy is definitely an important requirement, however.

apachano:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/GentenStudios/Phoenix/pullRequest?prid=4764724&bid=15884389)

M3TIOR:
@apachano @beeperdeeper089 get in here and review this shit lol. Marvel at the rediculous lengths I go to for ensuring documentation is up to par! :fist_raised: :headphones:

M3TIOR:
It's fucking perfect
evil laugh

Engine/Source/ImGuiHelpers.cpp
```diff
@@ -0,0 +1,274 @@
...
+	// 	renderLine(cache.substr(begin, current));
+
+	// For now this function actually handles newlines so it's a fix.
+	ImGui::TextWrapped("%s", (*text).c_str());
+};
+
+void BasicTerminal::drawOutputField(ImGuiWindowFlags flags)
+{
+	// May need later for text scrolling implementatinos.
+	// ImGuiContext* g = ImGui::GetCurrentContext();
```
DontBelieveMe: Instead of commenting it out we can just remove the comment and make note of it somewhere else if it's important.

Engine/Source/ImGuiHelpers.cpp
```diff
@@ -0,0 +1,274 @@
...
+		} while (cacheSize > targetOutputSize - hundredthSize);
+	}
+
+	// clear the output buffer and reset the cout internal object.
+	buf.clear();
+	cout.str(buf);
+}
+
+// NOTE: This inline is okay because it's only used within this file.
+inline void renderText(std::string* text)
```
DontBelieveMe: I'd mark this as `static` instead of `inline` - to avoid potential duplicate linker symbols.
Also instead of passing text as a pointer, I'd pass it as const ref just to be safe,

Engine/Source/ImGuiHelpers.cpp
```diff
@@ -0,0 +1,274 @@
...
+{
+	ImGuiWindow* window = ImGui::GetCurrentWindow();
+	// InputText is a frame object so we have to use push item width to adjust
+	// it.
+	// TODO: investigate label width changing or outright removal becaouse right
+	//   now we don't have a way to remove it, the data for it is just sitting
+	//   off screen to the right.
+	ImGui::PushItemWidth(ImGui::GetWindowWidth() -
+	                     (window->WindowPadding.x * 2.0f));
+	bool append_line = ImGui::InputText(
```
DontBelieveMe: This can be `const` and style conventions state this should be `camelCase` instead of `snake_case` - so `appendLine` in this case.

Engine/Source/ImGuiHelpers.cpp
```diff
@@ -0,0 +1,274 @@
...
+	    "\0",         // make sure we don't have any trailing text to the right.
+	    &inputBuffer, // Our output buffer object.
+
+	    // * These can be used in conjunction for extra functionality *
+	    // NOTE: The Callback doesn't need to be used for line by line input.
+	    ImGuiInputTextFlags(
+	        // Makes the widget return true on enter.
+	        ImGuiInputTextFlags_EnterReturnsTrue
+	        // ImGuiInputTextFlags_CallbackCompletion | // callback on tab
+	        // ImGuiInputTextFlags_CallbackHistory | // callback on up/down
```
DontBelieveMe: Remove instead of commenting out.

Engine/Source/ImGuiHelpers.cpp
```diff
@@ -0,0 +1,274 @@
...
+		{
+			(*cb)(inputBuffer.c_str(), cout);
+			// TODO:
+			//   Possibly implement a custom ostringstream for asynchonous data
+			//   transport between the calling functions and our flush method.
+			//   Might also be a good idea to combine the output streams of
+			//   all our child functions (which will probs be async eventually)
+			//   here.
+			flush();
+		}
```
DontBelieveMe: Could a range for be used here instead of accessing the iterators directly?

Sandbox/Main.cpp
```diff
@@ -1,11 +1,34 @@
...
 
+#include <iostream>
+#include <sstream>
 #include <memory>
+#include <string>
 
 using namespace q2;
 
+
+void rawEcho(const char* input, std::ostringstream &cout)
```
DontBelieveMe: Mark as static to avoid any nastly linker errors.

Sandbox/Main.cpp
```diff
@@ -1,11 +1,34 @@
...
+{
+	// easter egg commission for the tobster.
+	if (std::string(input).compare(std::string("buh-buh-bum-bah-bum")) == 0)
+		cout << "I'll be back\n";
```
DontBelieveMe: also, not a comment but just, lol

Engine/Include/Quartz2/ImGuiHelpers.hpp
```diff
@@ -0,0 +1,468 @@
...
+namespace q2
+{
+	/**
+	 * @namespace ImGuiHelpers
+	 * @brief
+	 *   This namespace serves as a ntainer for all ImGui extension materials
+	 *   maintained within Quartz/Phoenix including including but not limited to
+	 *   classes, functions, enums, etc.
+	 */
+	namespace ImGuiHelpers
```
DontBelieveMe: To match our project this namespace should probably be all lowercase, but if you're going for an extension like API for ImGui then I'd say its fine - it's up to you, just thought I should bring it up.

It'd also be acceptable to just use the `ImGui` namespace directly, which makes sense I think, seen as it's just helpers & extensions.

Engine/Include/Quartz2/ImGuiHelpers.hpp
```diff
@@ -0,0 +1,468 @@
...
+		 *   RecolorCW rcw("Chromotron");
+		 *   int main(){
+		 *     //Initialize ImGui, Setup frame loop...
+		 *     rcw.draw();
+		 *     cw.draw();
+		 *     //...
+		 *   }
+		 * @endcode
+		 */
+		class ImOOGuiWindow
```
DontBelieveMe: I'd probably call this `BaseImWindow` or something - the fact it's an object kindof implies the `OO` anyway.

Engine/Source/ImGuiHelpers.cpp
```diff
@@ -0,0 +1,274 @@
...
+	// std::string(data->Buf));
+	// comm_history.push_back()
+	return 1;
+}
+
+void BasicTerminal::flush()
+{
+	std::string buf       = cout.str();
+	int         bufSize   = buf.length();
+	int         cacheSize = cache.length();
```
DontBelieveMe: Same here - use `std::size_t` over `int` to match `cache.length()` type

Engine/Source/ImGuiHelpers.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+	// ceiling this so we don't run into an infinite loop with extremely small
+	// targetOutputSizes although it should never happen realistically because
+	// I made sure this is alligned in kilobytes of memory.
+	int hundredthSize = static_cast<int>(std::ceil(targetOutputSize / 100));
```
DontBelieveMe: Can be const and is it possible to make this name a bit more descriptive - hundrethSize of what?

Engine/Source/ImGuiHelpers.cpp
```diff
@@ -0,0 +1,274 @@
...
+	// Just for backup the while loop seeks a slightly larger segment
+	// so after we clear up
+	int tenthSize = static_cast<int>(std::floor(targetOutputSize / 10));
```
DontBelieveMe: Can be const and same with the name as above.

Engine/Source/ImGuiHelpers.cpp
```diff
@@ -0,0 +1,274 @@
...
+			// IDK if this is done programatically or not, best practice to do
+			// once, reuse return value since this is up to system
+			// implementation I believe.
+			cacheSize = cache.length();
+
+			// Auto cut at one hundredth the targetSize or newline, which ever
+			// it hits first.
+			for (int l = tenthSize; (l > 0) && cache[l] != '\n'; l++)
+			// for (auto l=0; (l < hundredthSize && cache[l] != "\n" && l <
+			// cacheSize; l++)
```
DontBelieveMe: This can be removed.

Sandbox/UI.hpp
```diff
@@ -0,0 +1,61 @@
...
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+#include <Quartz2/Quartz.hpp>
+#include <imgui.h>
+
+/// @brief Contains all UI elements specifically designed for Phoenix.
+namespace UI
```
DontBelieveMe: Namespace should be lowercase - `ui` in this case.

Sandbox/UI.hpp
```diff
@@ -0,0 +1,61 @@
...
+{
+	// @brief A restyle of the standard BasicTerminal for Phoenix
+	class ChatWindow : public q2::ImGuiHelpers::BasicTerminal
+	{
+		/// @privatesection
+	private:
+		/// @brief The current external focus event state.
+		bool renderFocus = false;
```
DontBelieveMe: Private member variables should have `m_` prefix.

Sandbox/UI.cpp
```diff
@@ -0,0 +1,154 @@
...
+#include <UI.hpp>
+
+// Standards
+//...
+
+using namespace UI;
+
+// hacked from the ImGui source
+static const float WINDOWS_MOUSE_WHEEL_SCROLL_LOCK_TIMER = 2.00f;
+void               ForceUpdateMouseWheel()
```
DontBelieveMe: This can be marked `static` and also doesn't appear to be used so can be removed (if you are 100% sure it's going to be used soon then that's ok, If you could just comment it to state as such)

Engine/Source/ImGuiHelpers.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+#include <imgui.h>
+#include <imgui_internal.h> // for nav inputs
+#include <misc/cpp/imgui_stdlib.h>
+
+using namespace q2::ImGuiHelpers;
+
+// This is a ImGuiInputTextCallback
+// TODO: add terminal history & hook for autocomplete.
+int callback(ImGuiInputTextCallbackData* data)
```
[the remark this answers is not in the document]
M3TIOR: That's a todo m8, I just left the scaffolding in there for when we implement the command recall and or completion hooks. That code is required for execution, and it's what we'll be modifying later.

Engine/Source/ImGuiHelpers.cpp
```diff
@@ -0,0 +1,274 @@
...
+	// 	break;
+	//
+	// 	// Filtered by imgui_stdlib implementation.
+	// 	// case ImGuiInputTextFlags_CallbackCharFilter:
+	// 	// 	std::cout << "CallbackCharFilter\n";
+	// 	// break;
+	// };
+	// comm_history.push_back(std::string("Contents: ") +
+	// std::string(data->Buf));
+	// comm_history.push_back()
```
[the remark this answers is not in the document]
M3TIOR: That's going to be used eventually just not atm. I'd rather leave it there then go back and snag it out of an old commit because that takes longer.

Engine/Source/ImGuiHelpers.cpp
```diff
@@ -0,0 +1,274 @@
...
+	return 1;
+}
+
+void BasicTerminal::flush()
+{
+	std::string buf       = cout.str();
+	int         bufSize   = buf.length();
```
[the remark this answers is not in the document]
M3TIOR: Fair.

Engine/Source/ImGuiHelpers.cpp
M3TIOR: Fair.

Engine/Source/ImGuiHelpers.cpp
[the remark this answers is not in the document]
DontBelieveMe: Yeah, but it can still be marked as `static` just to give it internal linkage save any nasty compile errors.

Engine/Include/Quartz2/ImGuiHelpers.hpp
```diff
@@ -0,0 +1,468 @@
...
+			 *  @brief
+			 *    The public rendering interface for subclasses.
+			 *
+			 *  @detailed
+			 *    Provides an interface for accessing drawEx that doesn't require
+			 *    the override cases be redefined in later classes. Be aware that
+			 *    if no drawEx funcion is defined in an inheriting class, this
+			 *    function will not work.
+			 */
+			void draw(bool* p_open, ImGuiWindowFlags flags);
```
beeperdeeper089: Since we call our other final render methods "render", perhaps we call it render instead of "draw"? (Not a formal request, just a question if it would work or sound weird)

Engine/Include/Quartz2/ImGuiHelpers.hpp
```diff
@@ -0,0 +1,468 @@
...
+			 *    standard conventions and must be passed to `begin` by
+			 *    the implementing class.
+			 *
+			 * @brief
+			 *   The public draw interface. This handles the actual rendering
+			 *   of the drawn class. This function must call `begin`, and `end`
+			 *   at the very least.
+			 *
+			 * @detailed
+			 *   This function implements the inherited classes rendering. It's
```
beeperdeeper089: Inherited **class'**, not classes :)

Engine/Include/Quartz2/ImGuiHelpers.hpp
```diff
@@ -0,0 +1,468 @@
...
+		};
+
+		/**
+		 * @typedef
+		 * @brief The callback function pointer format for all terminal callbacks.
+		 * @detailed
+		 *   Specifically, this is defined as an ease-of-use feature for definitions
+		 *   pertaining to the function callbacks within the `BasicTerminal` class.
+		 */
+		typedef void (*TerminalCallback)(const char* content, std::ostringstream &cout);
```
beeperdeeper089: Doing this has a chance of being misleading for some people, I see the intent, however would probably say typdeffing it to void isnt great, if you named it something else it would probably be fine.

This is in the q2 namespace (if mobile hasn't let me down again) so that could create issues if left as void

Sandbox/UI.cpp
```diff
@@ -0,0 +1,154 @@
...
+	static bool   focused;
+
+	/***
+	 * NOTE:
+	 *	`ImGui::PushStyleVar` & `PopStyleVar` for alpha is better unified
+	 *  across children than `ImGui::SetNextWindowBgAlpha` which sets each
+	 *  "Window's" transparency to a factor of the prior existing value for one
+	 *  element. `PushStyleVar` can apply to multiple elements once.
+	 */
+	if (focused)
```
beeperdeeper089: Use braces, we don't want an Apple happening to us.

M3TIOR: What do you mean?

Engine/Include/Quartz2/ImGuiHelpers.hpp
M3TIOR: Since this isn't the render loop, and drawing content I think changing it would be a bit confusing.

Engine/Include/Quartz2/ImGuiHelpers.hpp
M3TIOR: That sould be in the ImGuiHelpers namespace. And it's not typedefed to void, that's a function pointer definition that defines the type `TerminalCallback` to any instance of `void [fnName](const char*, std::ostringstream&);`

beeperdeeper089: Yeah fuck me, this is why i shouldn't review PRs in the morning.

M3TIOR: Nah, you're good. I will admit the syntax for function pointer definitions is a bit confusing.

Sandbox/UI.cpp
beeperdeeper089: do:
```
if (focused)
{
  blah...
}
```
instead of just
```
if (focused)
    blah...
```

The Apple thing I was talking about is [this](https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/).

M3TIOR: ... But :sob: it's so concise as it is. And because of clang format I'm not allowed to
```c++
if (something)
  { code(); }
```
:(

I'm sure with the right linter we should be able to prevent this anywho.

M3TIOR: I'll fix it :expressionless:. If I can find a better linter that can find when those problems exist, can we use that c++ feature again?

Engine/Source/ImGuiHelpers.cpp
DontBelieveMe: Yeah, the `inline` keyword is nothing but a very weak hint to the compiler to just substitute the function in place, as you said (although compilers basically just ignore them anyway, but that's not the point) - the point of using static here is that, if the compiler didn't inline it and just left it as a normal function, then it would add the function to the object files symbol table as per usual. This is OK until some other function in another object file comes along and has the same name - `renderText` and all of a sudden the linker has two functions both of the same name and then can't decide which one to use. So by marking this function as `static` this tells the compiler that we really won't use this function(or symbol) outside this translation unit so it does not need to add it to the symbol table and hence avoiding any duplicate symbols linker errors.
Yeah, we know as developers that logically this function is not used outside this file, but since in C++ functions have by default an external linkage (aka it has to be exported in the object file) the compiler doesn't know this.
Also by marking this static you could potentially enable even more optimisations than just inline would, without link time optimisation compilers cannot optimise across object files (for the most part).

TL;DR it can cause linker errors, compiler can probably optimise around it more and it's better practice.

Engine/Source/ImGuiHelpers.cpp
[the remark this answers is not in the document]
DontBelieveMe: Just for reference later, I've added a longish explanation comment below about the use of static

Engine/Source/ImGuiHelpers.cpp
DontBelieveMe: yeah I've used rather extensively - there are some comments as API documentation a bit over the place, but I know what you mean it has no actual proper documentation so you generally have t o look into the code quite a bit - which I guess is better in the long run, but short term it's a pain in the arse 😄 
In this case though the API is fairly clear as to what you can pass, and this commented out code is only listing enum values that you can pass, and that's easy to see anyway just by looking at the signature

Sandbox/Main.cpp
DontBelieveMe: 😄 yeah it's good, and on the linker/static note I added a longish comment on this above (copy pasted below just to make it easier to read)

---
Yeah, the inline keyword is nothing but a very weak hint to the compiler to just substitute the function in place, as you said (although compilers basically just ignore them anyway, but that's not the point) - the point of using static here is that, if the compiler didn't inline it and just left it as a normal function, then it would add the function to the object files symbol table as per usual. This is OK until some other function in another object file comes along and has the same name - renderText and all of a sudden the linker has two functions both of the same name and then can't decide which one to use. So by marking this function as static this tells the compiler that we really won't use this function(or symbol) outside this translation unit so it does not need to add it to the symbol table and hence avoiding any duplicate symbols linker errors.
Yeah, we know as developers that logically this function is not used outside this file, but since in C++ functions have by default an external linkage (aka it has to be exported in the object file) the compiler doesn't know this.
Also by marking this static you could potentially enable even more optimisations than just inline would, without link time optimisation compilers cannot optimise across object files (for the most part).

TL;DR it can cause linker errors, compiler can probably optimise around it more and it's better practice.

Engine/Source/ImGuiHelpers.cpp
M3TIOR: Thanks for the explanation. I actually didn't know cpp assumed external linkage by default so that's extremely helpful.

Sandbox/UI.cpp
beeperdeeper089: tbh I don't personally mind using just the no braces for a single line, but it's in our standards, and it is good practise to have braces, but depending on the actual context, we might be able to allow it.

On the linter issue, I think we should stick to clang-format due to it's actual widespread use and the fact that it's probably the best and most well-made option out there, and there almost definitely is a thing you can use in the settings file that should allow that. We can discuss this on discord if you find anything worth switching to if it's actually very good.

Engine/Include/Quartz2/ImGuiHelpers.hpp
M3TIOR: classes' < plural possession, almost got me lol.

Engine/Source/ImGuiHelpers.cpp
M3TIOR: This stuff is also going to be added in soon. Those pieces I know for sure will be used directly so I'm resolving this.

apachano:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/GentenStudios/Phoenix/pullRequest?prid=4764724&bid=15884389)

M3TIOR:
Calm your tits Codacy.

Sandbox/Main.cpp
```diff
@@ -1,11 +1,34 @@
...
+static void rawEcho(const char* input, std::ostringstream &cout)
+{
+	// easter egg commission for the tobster.
+	if (std::string(input).compare(std::string("buh-buh-bum-bah-bum")) == 0)
+		cout << "I'll be back\n";
+
+	cout << input << "\n";
+}
+
+ui::ChatWindow chat("Chat Window", 5,
```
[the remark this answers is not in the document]
M3TIOR: good to know

Engine/Include/Quartz2/ImGuiHelpers.hpp
```diff
@@ -0,0 +1,430 @@
...
+
+#include <iostream>
+#include <sstream>
+#include <string>
+#include <vector>
+
+// NOTE: consider transfering to an ANSI.h for submodule context, this code
+//   could possibly be used with an alternative GUI API as well and could
+//   benefit from being standalone.
+#define ANSI_RED IM_COL32(255, 0, 0, 255)
```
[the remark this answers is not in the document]
M3TIOR: I'm not sure yet whether or not I want to make a different file for that yet. I started doing that because I was having a distracting moment. I guess for now I'll remove it.

Engine/Source/ImGuiHelpers.cpp
```diff
@@ -0,0 +1,266 @@
...
+		while (outputSize > m_targetOutputSize - hundredthOutputSize);
+	}
+
+	// clear the output buffer and reset the cout internal object.
+	buf.clear();
+	cout.str(buf);
+}
+
+// NOTE: This inline is okay because it's only used within this file.
+static inline void renderText(std::string* text)
```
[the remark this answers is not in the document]
M3TIOR: Because I'm new to c++14 features and haven't looked up the documentation for it yet.

Engine/Source/ImGuiHelpers.cpp
```diff
@@ -0,0 +1,266 @@
...
+void BasicTerminal::drawInputField()
+{
+	ImGuiWindow* window = ImGui::GetCurrentWindow();
+	// InputText is a frame object, we have to use push item width to adjust it.
+	// TODO: investigate label width changing or outright removal becaouse right
+	//   now we don't have a way to remove it, the data for it is just sitting
+	//   off screen to the right.
+	ImGui::PushItemWidth(ImGui::GetWindowWidth() -
+	                     (window->WindowPadding.x * 2.0f));
+	bool appendLine = ImGui::InputText(
```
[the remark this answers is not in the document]
M3TIOR: Good catch, hard to remember because it's recycled every frame and const is only to make the rest of the code unable to modify it. It's syntax sugar / maintenance safety.

Sandbox/Main.cpp
```diff
@@ -1,11 +1,34 @@
...
 #include <memory>
+#include <string>
 
 using namespace q2;
 
+
+static void rawEcho(const char* input, std::ostringstream &cout)
+{
+	// easter egg commission for the tobster.
+	if (std::string(input).compare(std::string("buh-buh-bum-bah-bum")) == 0)
```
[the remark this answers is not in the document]
M3TIOR: std::strcmp is a c level funciton and not type safe. Why would I do that? Jk I don't feel like fixing this one because it will most likely be removed later.

Engine/Source/ImGuiHelpers.cpp
```diff
@@ -0,0 +1,266 @@
...
+	        ImGuiInputTextFlags_EnterReturnsTrue
+	        // ImGuiInputTextFlags_CallbackCompletion | // callback on tab
+	        // ImGuiInputTextFlags_CallbackHistory | // callback on up/down
+
+	        // Can't be used with cb_completion
+	        // ImGuiInputTextFlags_AllowTabInput
+	        ),
+	    &callback,
+
+	    NULL // Can be a user data object but unnecessary here.
```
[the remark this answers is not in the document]
M3TIOR: although nullptr is C++ type safe.

Engine/Include/Quartz2/ImGuiHelpers.inl
```diff
@@ -0,0 +1,93 @@
...
+	inline BasicTerminal::BasicTerminal(const char* name, int outputKiloBytes,
+	                                    const char* initialContents)
+			: BaseImWindow(name)
+	{
+		m_outputWindowName = std::string(name).append("_Output").c_str();
+		m_targetOutputSize = outputKiloBytes * 1024; // align to kilobytes
+		m_inputBuffer      = std::string("");
+		m_outputBuffer     = initialContents;
+		m_outputBuffer.reserve(m_targetOutputSize);
+		cout = std::ostringstream(std::ios_base::out | std::ios_base::ate);
```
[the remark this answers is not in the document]
M3TIOR: What?

M3TIOR:
Let's get this thing merged!

Engine/Include/Quartz2/ImGuiHelpers.inl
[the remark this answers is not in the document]
DontBelieveMe: https://www.learncpp.com/cpp-tutorial/8-5a-constructor-member-initializer-lists/

Engine/Source/ImGuiHelpers.cpp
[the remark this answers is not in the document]
DontBelieveMe: const reference is not C++14 (It's as old as C++ pretty much), since there doesn't appear to be a specific reason, use `const std::string& text` here instead of a string pointer.

(You may not need this, but just in case https://www.geeksforgeeks.org/references-in-c/)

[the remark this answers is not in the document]
M3TIOR: I already googled it.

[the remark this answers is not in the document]
M3TIOR: Eventually I'm going to go to the RFC / standardization submission for some better understanding.

Sandbox/Main.cpp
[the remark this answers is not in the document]
DontBelieveMe: Well if it's going to be committed to upstream then I'd rather it's done properly - you could say that all code is probably going to be removed so whats the point.

This is also going to be creating a load of unnecessary copies whenever `rawEcho` is called, if done thsi way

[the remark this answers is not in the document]
apachano: Still a little fuzzy on the whole thing but it looks like this will be called frequently?

If that is the case and this will be called frequently, even temporary code, I would go for what is going to be memory efficient especially if no major redesign needs to be done/ the change is straight forward. Just so it's not a todo later. We definitely can't make the most performant thing the first time we implement something (and shouldn't spend the time trying) but its worth catching the things we can.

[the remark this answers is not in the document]
M3TIOR: I wasn't expecting this to last the next few features / the first use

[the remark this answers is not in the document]
M3TIOR: I wasn't expecting this to last past the first use of the terminal... It's test code jargon.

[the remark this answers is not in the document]
M3TIOR: We don't even have to pass a function into the register callback method once, it just means the terminal won't do anything.

Engine/Include/Quartz2/ImGuiHelpers.inl
```diff
@@ -54,12 +54,12 @@ namespace ImGui
 
 	inline BasicTerminal::BasicTerminal(const char* name, int outputKiloBytes,
 	                                    const char* initialContents)
-			: BaseImWindow(name)
+			: BaseImWindow(name), m_inputBuffer(), m_outputBuffer(initialContents)
 	{
 		m_outputWindowName = std::string(name).append("_Output").c_str();
```
DontBelieveMe: I'm slightly worried that this is a bit broken - I'm thinking that you create a temporary `std::string` for `name` and append `"_Output"` to it - that's fine but you then store the C string pointer to that temporary string - so when the temporary `std::string` goes out of scope will then `m_outputWindowName` be invalidated? As the temporary std::string would have freed the memory it pointed to when it goes out of scope.

This could be solved by just storing m_outputWindowName as a `std::string` which would then automatically manage the memory for us.

Engine/Include/Quartz2/ImGuiHelpers.inl
```diff
@@ -54,12 +54,12 @@ namespace ImGui
...
-		m_targetOutputSize = outputKiloBytes * 1024; // align to kilobytes
-		m_inputBuffer      = std::string("");
-		m_outputBuffer     = initialContents;
+
+		// align to kilobytes
+		m_targetOutputSize = outputKiloBytes * 1024;
```
DontBelieveMe: This could go in the constructor initializer list.

Engine/Include/Quartz2/ImGuiHelpers.inl
```diff
@@ -54,12 +54,12 @@ namespace ImGui
...
 		m_outputBuffer.reserve(m_targetOutputSize);
 		cout = std::ostringstream(std::ios_base::out | std::ios_base::ate);
```
DontBelieveMe: In fact as could this, I'm pretty sure (could go in initializer list)

DontBelieveMe:
OK, so I know I said that'd be the last review... but I think I may have spotted something that could be fairly major, so better safe than sorry right

Engine/Include/Quartz2/ImGuiHelpers.inl
M3TIOR: Didn't know I could initialize any member using an initializer list but hey, apperently I've never really used C++ before so I'm learning something.

Engine/Include/Quartz2/ImGuiHelpers.inl
M3TIOR: Just found out ImGui does it's own Window Name "namespace" handling; which I was trying to do because it looked necessary. So I dropped the member entirely.

M3TIOR:
:sweat_smile: Wooooh

PR #48, opened 2020-01-09
Feat doxy

apachano:
Config file for Doxygen

If approved, feel free to merge


Doxyfile
```diff
@@ -0,0 +1,2517 @@
...
+# the logo to the output directory.
+
+PROJECT_LOGO           =
+
+# The OUTPUT_DIRECTORY tag is used to specify the (relative or absolute) path
+# into which the generated documentation will be written. If a relative path is
+# entered, it will be relative to the location where doxygen was started. If
+# left blank the current directory will be used.
+
+OUTPUT_DIRECTORY       =
```
beeperdeeper089: Set this to like ./Documentation/Build

We can put other MDs we need into Documentation as well if needed.

Doxyfile
```diff
@@ -0,0 +1,2517 @@
...
+
+# If you use STL classes (i.e. std::string, std::vector, etc.) but do not want
+# to include (a tag file for) the STL sources as input, then you should set this
+# tag to YES in order to let doxygen match functions declarations and
+# definitions whose arguments contain STL classes (e.g. func(std::string);
+# versus func(std::string) {}). This also make the inheritance and collaboration
+# diagrams that involve STL classes more complete and accurate.
+# The default value is: NO.
+
+BUILTIN_STL_SUPPORT    = NO
```
beeperdeeper089: Change to yes.

Doxyfile
```diff
@@ -0,0 +1,2517 @@
...
+
+# If the EXTRACT_ALL tag is set to YES, doxygen will assume all entities in
+# documentation are documented, even if no documentation was available. Private
+# class members and static file members will be hidden unless the
+# EXTRACT_PRIVATE respectively EXTRACT_STATIC tags are set to YES.
+# Note: This will also disable the warnings about undocumented members that are
+# normally produced when WARNINGS is set to YES.
+# The default value is: NO.
+
+EXTRACT_ALL            = NO
```
beeperdeeper089: We need to experiment with this, not all of our code is currently documented but if we can get it into doxygen regardless, that would be epic.

apachano: I like the idea of being able to get warnings, these settings impact using doxygen itself but not IDE internal doxy builders used in tooltips (something I very much so like). Not terribly worried at this point but at some point in the future, we should make an effort to use the warnings to put together a list of things that need to be documented.

apachano: In short, I heavily vote against enabling this in favor of completing the documentation.

beeperdeeper089: Aight, no problem.

PR #49, opened 2020-01-09
Add doxy comments for file and classes, update license to 2020

apachano:
Just some quick Doxy, no functional changes.

If approved feel free to merge.

### DIFF
--- a/Engine/Include/Quartz2/Settings.hpp
+++ b/Engine/Include/Quartz2/Settings.hpp
@@ -1,4 +1,4 @@
-// Copyright 2020 Genten Studios
+// Copyright 2019-20 Genten Studios
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions are met:
@@ -30,7 +30,7 @@
  * @file Settings.hpp
  * @brief Implements a settings management system
  * 
- * @copyright Copyright (c) 2020 Genten Studios
+ * @copyright Copyright (c) 2019-20 Genten Studios
  * 
  */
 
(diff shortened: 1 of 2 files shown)


beeperdeeper089:
Refer to my comment the issue you made for this (#50) let's see whether we actually need to do this.

apachano:
@beeperdeeper089 There's still some doxy in there but I updated to match what Toby did.

Engine/Include/Quartz2/Settings.hpp
```diff
@@ -26,6 +26,14 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
+/**
+ * @file Settings.hpp
+ * @brief Implements a settings management system
+ * 
+ * @copyright Copyright (c) 2019-20 Genten Studios
```
beeperdeeper089: We don't need the copyright twice in one file, the file header is enough I would say.

Engine/Include/Quartz2/Settings.hpp
```diff
@@ -51,6 +66,13 @@ namespace q2
 
 	public:
 		Setting();
+		/**
+		 * @brief Construct a new Setting object
+		 * 
+		 * @param name The human readable name for the setting
```
beeperdeeper089: In terms of syntax do we want to standardise using param[in] and param[out] and [in/out] or do we just wanna "ignore" them to some point?

PR #61, opened 2020-01-12
Feat filesystem

apachano:
Addresses: None
Authors:
@apachano 

## Summary of changes
-  Adds a filesystem header to use in place of include<filesystem> to support legacy machines
-  Adds a check before attempting to load Lua modules that creates a savefile so we don't error out
  
## Caveats
-  None

## On approval
Please merge


Engine/Include/Quartz2/CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ set(Headers
 	${currentDir}/ImGuiHelpers.hpp
 	${currentDir}/ContentLoader.hpp
 	${currentDir}/Settings.hpp
+	${currentDir}/FileSystem.hpp
```
M3TIOR: Whoops, the file name you specified in the CMakeLists.txt is different from the actual filename.

M3TIOR: Perhaps rename `Filesystem.hpp` to `FileSystem.hpp`

Engine/Include/Quartz2/Filesystem.hpp
```diff
@@ -0,0 +1,80 @@
+
+/**
+ * @brief std::filesystem feature support for old systems
+ * 
+ * This checks if the compiler supports std::filesystem and if not,
+ * sets std::filesystem to equal std::experimental::filesystem
+ * 
+ * This file is just copied from a stack overflow post
+ * https://stackoverflow.com/questions/53365538/how-to-determine-whether-to-use-filesystem-or-experimental-filesystem
```
beeperdeeper089: Replace that with: https://stackoverflow.com/a/53365539/8279989

That's the actual answer, not just the post, just in case.

CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
+set(LIBCXX_ENABLE_EXPERIMENTAL_LIBRARY ON)
```
M3TIOR: This should probably be protected by an OSX check conditional

M3TIOR: NVM I'm stupid

PR #62, opened 2020-01-12
Fixed filesystem build issues.

M3TIOR:
Also added Beeper's credit where necessary. Please review the statement made in my commit log.

### DIFF
--- a/Engine/Include/Quartz2/FileSystem.hpp
+++ b/Engine/Include/Quartz2/FileSystem.hpp
@@ -1,12 +1,12 @@
 
 /**
  * @brief std::filesystem feature support for old systems
- * 
+ *
  * This checks if the compiler supports std::filesystem and if not,
  * sets std::filesystem to equal std::experimental::filesystem
  * 
  * This file is just copied from a stack overflow post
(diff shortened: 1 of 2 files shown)


M3TIOR:
Looks to be a c++11 issue

M3TIOR:
https://stackoverflow.com/questions/33149878/experimentalfilesystem-linker-error

So useing `std::experimental::filesystem` is (most likely) impossible without the pre-existing build flag `-lstdc++fs` based on that source information. I have no way of testing this theory without pushing the changes upstream. Would someone else mind cloning my changes and making sure things build on your machine as well? It builds on mine. I need to fine a system that this doesn't build on for testing purposes. That or I could just keep shoving shit into Travis and have it complain all night.

M3TIOR:
But ironically since we're using C++17 standard in our code, we shouldn't be needing to use the `std::experimental:filesystem` at all.

apachano:
It compiles fine on my computer as well. The problem is OSX didn't add support for std::filesystem until 10.15 (Released this most recent November?) so in order to support OSX that isn't this up to date we need the experimental if compiled on a machine without the std::filesystem support. That being said my 2014 mac mini was eligible for the 10.15 upgrade so maybe we just tell people to upgrade and get over it? @beeperdeeper089 What we are trying to do should be possible but it is worth it?

PR #64, opened 2020-01-13
Block + Texture Registry

beeperdeeper089:
Addresses: None
Authors:
@beeperdeeper089  

## Summary of changes
  - Newer block registry system
  - Texture handling system, no more duplicates. simpler system
  
## Caveats
  - There is a soft limit of 256 textures, more than this is UB since each graphics card supports a different amount of layers within a texture array.

## On approval
Please merge

### DIFF
--- a/Engine/Include/Quartz2/Voxels/Block.hpp
+++ b/Engine/Include/Quartz2/Voxels/Block.hpp
@@ -1,0 +1,83 @@
+// Copyright 2019 Genten Studios
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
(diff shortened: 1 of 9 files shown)


Engine/Include/Quartz2/Voxels/Block.hpp
```diff
@@ -0,0 +1,83 @@
...
+		public:
+			BlockType()  = default;
+			~BlockType() = default;
+
+			std::string displayName;
+			std::string id;
+
+			BlockCategory category = BlockCategory::AIR;
+
+			// top, left, back, right, top, bottom
```
apachano: Front?

beeperdeeper089: ah yeah

PR #72, opened 2020-01-15
Implement system to create save directory if one does not exist

apachano:
Addresses: none
Authors:
@apachano 

## Summary of changes
-  Creates a save folder when the program is run if one does not already exist. 
  
## Caveats
-  I mean, let's face it, this entire thing?

## On approval
Please merge

### DIFF
--- a/Phoenix/Source/Main.cpp
+++ b/Phoenix/Source/Main.cpp
@@ -81,7 +81,7 @@ static ui::ChatWindow chat("Chat Window", 5,
 	void initFiles(std::string save)
 	{
 		std::ofstream saveFile;
-		saveFile.open("Save/" + save + "/mods.txt")
+		saveFile.open("Save/" + save + "/mods.txt");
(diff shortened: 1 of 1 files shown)


Phoenix/Source/Main.cpp
```diff
@@ -53,6 +54,47 @@ static void rawEcho(const std::string& input, std::ostringstream& cout)
...
+	void initFiles(std::string save)
+	{
+		std::ofstream saveFile;
+		saveFile.open("Save/" + save + "/mods.txt");
+		if(saveFile != NULL){
+			saveFile.close();
+			return;
+		} else {
+			mkdir("Save", S_IRWXO);
+			mkdir("Save/" + save, S_IRWXO);
```
beeperdeeper089: do a ``.c_str()`` here to get the C string.

https://en.cppreference.com/w/cpp/string/basic_string/c_str

PR #73, opened 2020-01-15
Set up CI with Azure Pipelines

beeperdeeper089:


M3TIOR:
@beeperdeeper089 Is it possible that you could expain a little better in your commit messages why you need to do what looks like 10 redundant commits?

beeperdeeper089:
Idek what I'm doing myself, I'm gonna have to squash and merge this anyway so it's only one commit

M3TIOR:
I just thought it could be nice to be able to learn from your mistakes as well as my own. That way whatever you're getting wrong is more preventable in the future.

apachano:
@beeperdeeper089 Besides the mess where other branches got merged in, this looks good. My only question is what is the latest version of OSX it supports?

beeperdeeper089:
Mojave 10.14 if that means anything to you

https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops

apachano:
It wont compile filesystem :/ and depending on the xtools version it will still set all the flags to true so we cant detect that it cant until it throws a custom error.

apachano:
Same as Travis

beeperdeeper089:
We could maybe install a custom version of clang?

What we could do is build CLang for macOS and store it in the Azure Artifacts and use that build to compile? That's long and winded so preferably not but yeah, we need to figure out something for Crapple.

PR #76, opened 2020-01-18
refactor(feat-command): Now uses appropriate headers.

M3TIOR:


apachano:
My only concern is why are we using istream and ostream? Is there any sort of benefit to defining them separately? I start to become hesitant when we start blindly letting computers make decisions for us.

M3TIOR:
It's because you only use `istream` and `ostream` objects explicitly in the post method. The `<iostream>` header includes both of those and defines the conjoined `iostream` type as well. By including each separately, you remove the extra symbol bloat introduced by the `iostream` object. It's only a few bytes so it's negligable IMHO but it's still something.

M3TIOR:
For reference [the io documentation](https://en.cppreference.com/w/cpp/io) has the headers marked above each of the corresponding objects.

PR #77, opened 2020-01-19
Feat lua

apachano:
Addresses: None
Authors:
@apachano 

## Summary of changes
- Add default Modules and Save file to repo + have cmake copy on build
- Change the copy method for assets to work on all systems
- Add block registration function
- Modify module loading to be a class + keep track of what mod is loading when accessing assets
  
## Caveats
-  The executable is still not portable

## On approval
Please merge


apachano:
Since Assets/ needs to be copied every time we clone, it should be fine for the Save to do the same. This gets us loading content via Lua.

Phoenix/CMakeLists.txt
```diff
@@ -15,11 +14,17 @@ set_target_properties(${PROJECT_NAME} PROPERTIES
...
 #################################################
 
 source_group(TREE "${CMAKE_CURRENT_SOURCE_DIR}/Include/Phoenix" PREFIX "Header Files" FILES ${Headers})
 source_group(TREE "${CMAKE_CURRENT_SOURCE_DIR}/Source" PREFIX "Source Files" FILES ${Sources})
+
+#################################################
+## COPY SAVE, ASSETS, and MODULES TO BUILD DIR ##
+#################################################
+
+file(COPY Save DESTINATION ${CMAKE_BINARY_DIR}/Phoenix)
```
beeperdeeper089: Nice, didn't know you could do that.

Phoenix/Include/Phoenix/ContentLoader.hpp
```diff
@@ -78,15 +74,15 @@ namespace phx
 		 */
 		static bool loadModules(std::string save, sol::state& lua);
```
beeperdeeper089: You should probably do ``const std::string&``

Phoenix/Source/ContentLoader.cpp
```diff
@@ -53,7 +53,7 @@ Mod::Mod(std::string modName) : name(std::move(modName))
 
 Mod::~Mod() {};
```
beeperdeeper089: Use ``= default`` for this, the compiler can optimise as necessary. (i think, but either way, do that instead of an empty destructor)

Phoenix/Source/ContentLoader.cpp
```diff
@@ -103,6 +103,7 @@ bool modules::loadModules(std::string save, sol::state& lua)
 			// list Otherwise, move mod to back of load queue
 			if (satisfied)
 			{
+				m_currentMod = mod.name;
 				lua.script_file("Modules/" + mod.name + "/Init.lua");
```
beeperdeeper089: What happens if an error occurs here?

Phoenix/Source/Main.cpp
```diff
@@ -98,37 +98,36 @@ class Phoenix : public events::IEventListener
...
-				    "Assets/grass_side.png", "Assets/grass_side.png",
-				    "Assets/grass_top.png",  "Assets/dirt.png",
-				};
-			}
-
-			BlockRegistry::get()->registerBlock(grassBlock);
+			m_window->close();
 		}
 
+		// {
```
beeperdeeper089: Just dump this, we'll find it in git history if we need it.

Phoenix/CMakeLists.txt
apachano: I think so? if you do `cmake -S.` it will for sure but it's inconsistent with the regular build command. 

TBH the old one didn't work on Mac at all, I still manually copied the assets dir every time, cant speak for linux.

Phoenix/Include/Phoenix/ContentLoader.hpp
apachano: Call out issues with code I didn't change will ya :P
JK easy fix

Phoenix/Source/ContentLoader.cpp
apachano: If Sol catches it, it will vomit why to the terminal then terminate the program. 

We can implement better handling and output cleaner errors into a window, but we need a proper plan to handle errors (likely a system that will also handle displaying the error message to the user, log wtf happened, etc). Something for the future

Phoenix/Include/Phoenix/ContentLoader.hpp
apachano: Call out issues with code I didn't change will ya :P
JK easy fix

apachano: Call out issues with code I didn't change will ya :P
JK easy fix

apachano: Call out issues with code I didn't change will ya :P
JK easy fix

apachano:
@beeperdeeper089 REady for re-review

apachano:
@beeperdeeper089 Okay this fully implements callbacks as well so it's ready for review.

Phoenix/Include/Phoenix/Voxels/Block.hpp
```diff
@@ -44,6 +45,8 @@ namespace phx
 			LIQUID
 		};
 
+		typedef std::function<void(math::vec3 pos)> BlockCallback;
```
beeperdeeper089: do ``using BlockCallback = std::function<void(math::vec3 pos)>;``

It just looks nicer and is more modern.

apachano: Gross, modern

apachano:
@beeperdeeper089 Fixed, just waiting on CI

tobyplowy:
Nice =D

PR #80, opened 2020-01-20
Feat lua commander

apachano:
Authors:
@apachano 
@M3TIOR 

## Summary of changes
- Implements a Lua API for the commander
    - Lua files can now register a command
    - Lua files can now output text to the terminal

## On approval
Please merge


Phoenix/Include/Phoenix/ContentLoader.hpp
```diff
@@ -64,23 +67,23 @@ namespace phx
...
 		//TODO : Add proper error handling instead of returning a boolean
 
 		/**
 		 * @brief Loads the Lua API for use in modules
-		 * 
+		 *
 		 * @param lua The sol state used during runtime
 		 */
-        static void loadAPI(sol::state& lua);
+        static void loadAPI(sol::state& lua, ImGui::BasicTerminal& chat);
```
beeperdeeper089: Do we really *need* to pass a chat in here? Can the chat just not request a command book or whatever?

apachano: Since the chat terminal itself doesn't have a registry or global accessor we do. I could probably build a cleaner workaround but this was the simplest way to access the output stream attached to the chat terminal. 

The output to the chat terminal will likely have two additional iterations in the future. The first will be when I set the API up so you can register a function from anywhere (still deciding if a full registry or just an accessor to the sol object is appropriate) which will resolve this issue. The second will overhaul this API component when we implement multiplayer and need to determine which players terminal we are sending the message to.

PR #84, opened 2020-01-26
fix(feat-filesystem): Corrected issue with FindFilesystem.cmake INTERFACE

M3TIOR:
Horray! We're now configured to use the experimental filesystem when the final doesn't exist. Will fail if neither exist.

### DIFF
--- a/Engine/CMakeLists.txt
+++ b/Engine/CMakeLists.txt
@@ -4,7 +4,9 @@ add_subdirectory(ThirdParty)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_LIST_DIR}/Tools/CMake")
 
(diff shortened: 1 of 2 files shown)


SuperFola:
Please see the checks output to fix your code (eg. https://travis-ci.com/GentenStudios/Phoenix/jobs/280140401#L542)

beeperdeeper089:
Do we have any progress on this? We don't currently need it that badly, it's functional enough without it, so we can leave this for now if we don't actually need it for core functionality. We can sort this out once we actually plan on building the engine and providing filesystem functionality within there.

beeperdeeper089:
You will also need to merge in develop and stuff to get this working, and that will be painful so if you plan on continuing this, just make a new PR.

M3TIOR:
Yeah I'm abandoning this.

Issue #109, opened 2020-02-04
Crash when breaking last block in a chunk


tobyplowy:
## Describe the bug
I get a crash when I break all of the grass blocks in a chunk.
As soon as the last block is broken it crashes. #104 May be related to this.
Doing this also gives me an exception message tho it does not seem to be the same as #104

The message that I get for this crash.
Exception thrown at 0x08DBFE35 in Phoenix.exe: 0xC0000005: Access violation reading location 0x00000000.

## To Reproduce
1. Remove all blocks in a chunk by hand. (Tip use the chunk walls as a guide)
2. Remove the last remaining block of the chunk.
3. Crash!

## Expected behavior
It should not crash when you remove all blocks in a chunk.

## Additional context
Yes I actually removed all blocks in a chunk 4 times now. BY HAND! And all of that just to debug =D
I love this stuff (Not joking)

beeperdeeper089:
That is very funky. I'm not sure why this would even be an issue, thanks for the report!

Issue #108, opened 2020-02-04
You can place blocks on the corners (and diagonally) of other blocks


tobyplowy:
## Describe the bug
It's possible to place blocks diagonally on to blocks and you can also place them on corners of blocks. (Check demo video below)

## To Reproduce
1. Aim at a corner or side intersection of a block.
2. Place a block. (It may take multiple attempts to get the bug to happen. Just move the mouse a bit and try again)

## Expected behavior
You should only be able to place blocks on faces of other blocks, not on verts or lines.

## Additional context
A video showing the bug in action https://youtu.be/4XmWMify8vE **(the bug happens at 0:11)**

Issue #105, opened 2020-02-03
Block culling is broken and blocks are being removed.


tobyplowy:
## Describe the bug
When you place about 10 or so blocks above the grass blocks there start to be culling issues (Faces disappear for no reason as there is no block that is culling them or it may have become invisible) 
(This only happens above the grass chunks not inside them.)

## To Reproduce
1. Place about 10 or so blocks next/on top of each other (Make sure that you are above the grass chunks that generate)
2. Keep placing until you notice faces and blocks disappearing.

## Additional context
Here's a video of it in action. https://youtu.be/FoCXaqRNIl4?t=6

Issue #104, opened 2020-02-03
Crash when breaking a block outside grass chunk.


tobyplowy:
## Describe the bug
I get a crash when breaking a block in the area above the grass chunk. (Exception massage below) It does not happen if you remove blocks inside the grass chunk itself. (Even when placed)
(If you place and remove multiple blocks it's fine just don't remove the first block that you place as that is the one that crashes.)

## To Reproduce
1. Place a block in the air area (Above the grass)
2. Remove the block and you get an Instant crash (It only works with the first block that you place if you place multiple blocks)

## Additional context
Exception thrown at 0x08B9FE35 in Phoenix.exe: 0xC0000005: Access violation reading location 0x00000000.
And this may also be useful?
![image](https://user-images.githubusercontent.com/16853304/73620070-7b589980-4630-11ea-90c0-cd454b1d4799.png)

apachano:
Hmm, :/ let's try to get repro steps, otherwise, there isn't anything we can really do to fix this.

tobyplowy:
Got it again. I think it happened because of the missing texture block.

tobyplowy:
Here's a video of me performing the crash.
Again this happens at complete random. (Sometimes it crashes and sometimes it does not)
https://youtu.be/oIWXKPQ3gcg

tobyplowy:
Now it seems to always happen...

tobyplowy:
Oh it only seems to happen if the block is outside (on top of) the grass chunk (In the air area)

tobyplowy:
@apachano I found the problem and I updated my issue to be more descriptive.

tobyplowy:
If you place multiple blocks and remove them it's fine but as soon as you remove the first block that you placed it crashes. So it's only the first block that you place that crashes.

beeperdeeper089:
That is one funky bug...

tobyplowy:
It sure is. Very interesting and specific.

tobyplowy:
#109 May be related to this crash.

PR #110, opened 2020-02-05
Feature: Simple Block Placing & Breaking.

beeperdeeper089:
Addresses: #109, #105, #104
Authors:
@beeperdeeper089 
@apachano 

## Summary of changes
-  Implement simple placing/breaking functionality.
-  Implement new Actor/Player system.
  
## Caveats
-  Blocks can be placed diagonally, no "face" recognition for application. (#108)

## On approval
Please wait for personal confirmation that this is ready to merge, if not commented or edited to reflect.

### DIFF
--- a/Phoenix/Include/Phoenix/Player.hpp
+++ b/Phoenix/Include/Phoenix/Player.hpp
@@ -58,9 +58,12 @@ namespace phx
 		bool action1();
 		bool action2();
 
+        void setHand(voxels::BlockType* block);
+        voxels::BlockType* getHand();
 
 	private:
 		float m_reach = 32.f;
 		voxels::ChunkManager* m_world;
+        voxels::BlockType* m_hand;
 	};
 } // namespace phx
--- a/Phoenix/Include/Phoenix/Player.hpp
+++ b/Phoenix/Include/Phoenix/Player.hpp
@@ -1,4 +1,4 @@
-// Copyright 2019-20 Genten Studios
+// Copyright 2019 Genten Studios
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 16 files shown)


Phoenix/Include/Phoenix/Player.hpp
```diff
@@ -0,0 +1,69 @@
...
+		bool action1();
+		bool action2();
+
+        void setHand(voxels::BlockType* block);
+        voxels::BlockType* getHand();
+
+	private:
+		float m_reach = 32.f;
+		voxels::ChunkManager* m_world;
+        voxels::BlockType* m_hand;
```
SuperFola: We should consider having a collection here, if we want to handle blocks in left and right hand maybe (or even a block on your head, who knows)

Phoenix/Source/Main.cpp
```diff
@@ -120,20 +148,29 @@ class Phoenix : public events::IEventListener
 
 		Settings::get()->load();
 
+		m_world  = new voxels::ChunkManager(3);
+		m_player = new Player(m_world);
+		m_camera = new gfx::FPSCamera(m_window);
```
SuperFola: Yes (dynamic allocation) but no (new is bad), should consider using either unique_ptr (no copy allowed) or shared ptr (copy allowed, handling multiple ownership)

Phoenix/Source/Player.cpp
```diff
@@ -0,0 +1,118 @@
...
+	{
+		pos.floor();
+
+		if (m_world->getBlockAt(pos)->category == voxels::BlockCategory::SOLID)
+		{
+			m_world->setBlockAt(
+			    pos, voxels::BlockRegistry::get()->getFromID("core:air"));
+
+			return true;
+		}
```
SuperFola: How are we going to handle liquids?

beeperdeeper089: We aren't bothering with liquids yet, we have no internal implementation and liquids are not planned till a later release.

Phoenix/Include/Phoenix/Player.hpp
beeperdeeper089: We will be doing that, this is a very primitive implementation, hence the "Simple" in the title. This however, is not a feature we've classed as being worth implementing within the scope of this PR, purely due to that not being necessary for the "Playable" aspect.

We're just trying to get enough code in to make sure we're making progress, obviously, more features will be added as time progresses.

Phoenix/Source/Main.cpp
beeperdeeper089: Since we're managing and we know the precise scope of these variables, I think using ``new`` should be fine, feel free to comment if you think otherwise.

Shared Pointers and Unique Pointers where unnecessary (like this use case)will just clutter code IMO - again feel free to comment otherwise.

SuperFola: Intelligent pointers are managing allocation, error handling on creation, and memory freeing on their own, so we would avoid using `delete`s and have a safer code, but I agree on the part where we know the precise scope of the variable.

I mentionned this because using `new` in C++14/17 is now seen as a bad practice

beeperdeeper089: Yeah I think we should definitely consider these elsewhere in the project since we rarely always no the exactly scope, but for this case, I think it wouldn't be seen as bad practice but as deliberate use since we know our exact scope.

If we don't use smart pointers for this, we reduce parameter bloat where we have to type out things like ``std::unique_ptr<blah> blah`` which can get very long in some cases.

While using smart pointers are good, I would argue that only using them where necessary is "good practice" since you're not adding unnecessary (even if negligible) overhead in situations where you know you're in control.

PR #123, opened 2020-02-11
Feat saving

### LINKED ISSUE #33: Chunk Loading/ Saving
## Story
We need to be able to load chunk data from a save file as well as save chunk data to a save file.

## MVP
What are the minimum features needed for this story?
- [ ] Function on the chunk class to convert a chunk into a string to be saved to a file
- [ ] Constructor for the chunk class to convert a string from a save file into a chunk

apachano:
Resolves #33 
Authors:
@apachano 

## Summary of changes
-  Implements a basic map system
-  Implements function and constructor to convert chunk to string
  
## Caveats
-  Saving is not optimized
-  The map does not unload chunks

## On approval
Please merge

### DIFF
--- a/Phoenix/Include/Phoenix/Math/Vector3.hpp
+++ b/Phoenix/Include/Phoenix/Math/Vector3.hpp
@@ -29,8 +29,8 @@
 #pragma once
 
 #include <cmath>
-#include <ostream>
 #include <iostream>
+#include <ostream>
 
 namespace phx::math
 {
@@ -338,33 +338,39 @@ namespace phx::math
 		}
 	};
 
-    // 'KeyEqual' parameter to use when using Vector3 inside a hash map.
-    struct Vector3KeyComparator
-    {
-        template <typename T>
-        bool operator()(const detail::Vector3<T>& a,
-                        const detail::Vector3<T>& b) const
-        {
-            return a.x == b.x && a.y == b.y && a.z == b.z;
-        }
-    };
+	// 'KeyEqual' parameter to use when using Vector3 inside a hash map.
+	struct Vector3KeyComparator
+	{
+		template <typename T>
+		bool operator()(const detail::Vector3<T>& a,
+		                const detail::Vector3<T>& b) const
+		{
+			return a.x == b.x && a.y == b.y && a.z == b.z;
+		}
+	};
 
-    // 'Key' parameter to use when using Vector3 inside a map.
-    struct Vector3Key
-    {
-        template <typename T>
-        bool operator()(const detail::Vector3<T>& a,
-                        const detail::Vector3<T>& b) const
-        {
-            if(a.x != b.x){return a.x < b.x;}
-            if(a.y != b.y){return a.y < b.y;}
-            if(a.z != b.z){return a.z < b.z;}
-            return false;
-        }
-    };
-
-
-
+	// 'Key' parameter to use when using Vector3 inside a map.
+	struct Vector3Key
+	{
+		template <typename T>
+		bool operator()(const detail::Vector3<T>& a,
+		                const detail::Vector3<T>& b) const
+		{
+			if (a.x != b.x)
+			{
+				return a.x < b.x;
+			}
+			if (a.y != b.y)
+			{
+				return a.y < b.y;
+			}
+			if (a.z != b.z)
+			{
+				return a.z < b.z;
+			}
+			return false;
+		}
+	};
 } // namespace phx::math
 
 template <typename T>
--- a/Phoenix/Include/Phoenix/Math/Vector3.hpp
+++ b/Phoenix/Include/Phoenix/Math/Vector3.hpp
@@ -199,15 +199,15 @@ namespace phx
 				}
 				bool operator!=(const Vector3& rhs) const
 				{
-					return (x != rhs.x && y != rhs.y && z != rhs.z);
+					return (x != rhs.x || y != rhs.y || z != rhs.z);
 				}
 				bool operator>(const Vector3& rhs) const
 				{
-					return (x > rhs.x && y > rhs.y && z > rhs.z);
+					return (x > rhs.x || y > rhs.y || z > rhs.z);
 				}
 				bool operator<(const Vector3& rhs) const
 				{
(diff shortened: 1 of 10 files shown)


Phoenix/Include/Phoenix/Voxels/ChunkManager.hpp
```diff
@@ -74,20 +73,22 @@ namespace phx
 			 * @brief Constructs the ChunkManager
 			 * @param viewDistance The view distance in every direction.
 			 */
-			ChunkManager(int viewDistance);
+			ChunkManager(int viewDistance, Map map);
```
SuperFola: maybe it would be better to take a const Map&

Phoenix/Include/Phoenix/Voxels/Map.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include <Phoenix/Voxels/Chunk.hpp>
+#include <map>
+#include <Phoenix/Math/Math.hpp>
+
+namespace phx::voxels
+{
+    class Map{
+    public:
+        Map(std::string save, std::string name);
```
SuperFola: for memory usage, better to take const std::string&

Phoenix/Include/Phoenix/Voxels/Map.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+        Chunk getChunk(math::vec3 pos);
+        void setBlockAt(math::vec3 pos, BlockType* block);
+        void save(math::vec3 pos);
```
SuperFola: would need to check but may be better to have const math::vec3&

beeperdeeper089: math::vec3 should be fine, its literally only 3 floats

Phoenix/Include/Phoenix/Voxels/Map.hpp
apachano: I move the memory in the constructor instead

apachano:
@beeperdeeper089 Can you take a look at this when you get a sec.

Phoenix/Include/Phoenix/Math/Vector3.hpp
```diff
@@ -30,327 +30,342 @@
...
+        bool operator()(const detail::Vector3<T>& a,
+                        const detail::Vector3<T>& b) const
+        {
+            if(a.x != b.x){return a.x < b.x;}
+            if(a.y != b.y){return a.y < b.y;}
+            if(a.z != b.z){return a.z < b.z;}
+            return false;
+        }
+    };
+
```
beeperdeeper089: Empty newlines?

Phoenix/Include/Phoenix/Voxels/Chunk.hpp
```diff
@@ -34,138 +34,139 @@
 
 #include <vector>
 
-namespace phx
+namespace phx::voxels
```
beeperdeeper089: Why are we doing this?

Phoenix/Include/Phoenix/Math/Vector3.hpp
```diff
@@ -30,327 +30,342 @@
 
 #include <cmath>
 #include <ostream>
+#include <iostream>
 
-namespace phx
+namespace phx::math
```
beeperdeeper089: Why are we doing this? (Just curious, idm which one we use but we have to standardise it now or keep it as it was.)

Phoenix/Include/Phoenix/Math/Vector3.hpp
```diff
@@ -30,327 +30,342 @@
...
+        template <typename T>
+        bool operator()(const detail::Vector3<T>& a,
+                        const detail::Vector3<T>& b) const
+        {
+            return a.x == b.x && a.y == b.y && a.z == b.z;
+        }
+    };
+
+    // 'Key' parameter to use when using Vector3 inside a map.
+    struct Vector3Key
```
beeperdeeper089: If this works, delete the other one and rename this to that - so it makes more sense.

Phoenix/Include/Phoenix/Voxels/Chunk.hpp
```diff
@@ -34,138 +34,139 @@
...
 {
-	namespace voxels
```
beeperdeeper089: Why is this even happening?

Phoenix/Include/Phoenix/Voxels/Map.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+#pragma once
+
+#include <Phoenix/Voxels/Chunk.hpp>
+#include <map>
+#include <Phoenix/Math/Math.hpp>
+
+namespace phx::voxels
+{
+    class Map{
```
beeperdeeper089: Clang format.

Phoenix/Include/Phoenix/Math/Vector3.hpp
apachano: CLion suggested it. IMO it reduces lines and indents without making the code harder to read and it's easier to read due to more code fitting on the screen.

Unless you have any objections, I think we should standardize following this method.

I'm hesitant to spend a ton of time fixing old code but we can make it a point to clean up as we move forward like we are with Doxygen.

Phoenix/Include/Phoenix/Math/Vector3.hpp
apachano: You use the other one somewhere in the renderer, my first attempt to do this fix was re-writing the original but it broke things.

Phoenix/Include/Phoenix/Voxels/Chunk.hpp
apachano: CLion automates it

apachano:
@beeperdeeper089 Replied to your comments, I had CLion format all the files I touched.

Phoenix/Include/Phoenix/Math/Vector3.hpp
beeperdeeper089: CLion should be able to do this in one command, take a look if it can, otherwise this is fine as it is.

PR #124, opened 2020-02-13
Add information around how to contribute

apachano:
Adds basic information on how to get involved in the project.


tobyplowy:
Oh you removed my review?
I guess you can ignore my review then.

apachano:
I didn't think your review would be enough but it looks like it is. Merging now.

Issue #162, opened 2020-03-24
Implement ENTT Third Party


apachano:
## Story
We want to use ENTT as we work in the future. The first step is to add it as a submodule in the third parties and get CMake recognizing it. 

## MVP
What are the minimum features needed for this story?
-  [x] ENTT is added as submodule
-  [x] CMake recognizes ENTT
-  [x] ENTT can be used and any setup that needs to happen to use it is done.

PR #164, opened 2020-03-25
Feat entt

SuperFola:
Resolves: #162
Authors:
@SuperFola 

## Summary of changes
-  adding entt as a submodule
- referecing the entt/src directory as an include directory (this isn't a mistake, the lib is 100% header only)
- adding Sandbox/ and testing entt inside it
  
## Caveats
-  no

## On approval
Please merge


beeperdeeper089:
There's no real point of having the sandbox there, just shove like tests into the Server or something for now since there's nothing there. It just adds a second folder that we don't actually need.

PR #169, opened 2020-04-05
Feat actor

apachano:
Authors:
@apachano 

## Summary of changes
-  Broke Actor class out into an ECS model
  
## Caveats
-  None

## On approval
Please merge


Phoenix/Client/Include/Client/Graphics/Camera.hpp
```diff
@@ -204,7 +205,7 @@ namespace phx::gfx
 		 * @brief Sets the actor for the camera to "follow".
 		 * @param actor The actor to set the camera to use.
 		 */
-		void setActor(Actor* actor);
+		void setActor(entt::entity actor);
```
SuperFola, line 208: Why aren't you taking a reference here? The registry wouldn't be modified

apachano: The EnTT registry accessors won't accept a reference to an entity identifier. We need to pass the actual thing for it to work.

PR #183, opened 2020-04-20
Feat enet queue

apachano:
Authors:
@apachano 

## Summary of changes
-  Adds a queuing system and builds that system for state processing
-  Processes queued states in their sequence order
  
## Caveats
-  This only works with one user, we still need to capture how many users we are expecting packets from.

### DIFF
--- a/Phoenix/Server/Source/Game.cpp
+++ b/Phoenix/Server/Source/Game.cpp
@@ -62,6 +62,8 @@ void Game::run()
 				ActorSystem::tick(m_registry,
 				                  m_registry->get<Player>(*state.first).actor,
 				                  dt, state.second);
+
+				// @todo remove this debug statement before merging to develop
 				std::cout
 				    << m_registry
 				           ->get<Position>(
--- a/Phoenix/Server/Source/Game.cpp
+++ b/Phoenix/Server/Source/Game.cpp
@@ -26,40 +26,39 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
-#pragma once
-
 #include <Server/Game.hpp>
-#include <Server/Iris.hpp>
-
 #include <Server/User.hpp>
 
-//#include <Server/Commander.hpp>
+#include <Common/Actor.hpp>
+#include <Common/Movement.hpp>
+#include <Common/Position.hpp>
 
-#include <entt/entt.hpp>
-#include <enet/enet.h>
+using namespace phx;
+using namespace phx::server;
 
-#include <array>
-#include <string>
-
-namespace phx::server
+Game::Game(entt::registry* registry, bool* running, networking::Iris* iris)
+    : m_registry(registry), m_running(running), m_iris(iris)
 {
+}
 
-	class Server
+void Game::run()
+{
+	while (m_running)
 	{
-	public:
-		Server(std::string save);
-		~Server();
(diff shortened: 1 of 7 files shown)


Phoenix/Server/Source/Iris.cpp
```diff
@@ -70,59 +69,57 @@ Iris::~Iris() { enet_host_destroy(m_server); }
...
 				break;
 			}
+
+			/* Clean up the packet now that we're done using it. */
+			enet_packet_destroy(m_event.packet);
+			break;
+
+		case ENET_EVENT_TYPE_DISCONNECT:
+			printf("%s disconnected.\n",
+			       reinterpret_cast<const char*>(m_event.peer->data));
```
beeperdeeper089, line 118: Areyou sure we need a ``reinterpret_cast`` here? They can be dangerous, if we can stick with ``static_cast<>`` that would be best, otherwise it's fine since we should be able to guarantee safety.

M3TIOR: I agree, when researching this `static_cast` was my thought as well because reinterpret cast can mutate the data beneath the variable while `void*` conversions are handled implicitly by C++ so `static_cast` can do it without data mutation.

Phoenix/Server/Source/Game.cpp
```diff
@@ -0,0 +1,75 @@
...
+		else
+		{
+			printf("State");
+			networking::StateBundle m_currentState = m_iris->stateQueue.front();
+			m_iris->stateQueue.pop_front();
+
+			for (auto state : m_currentState.states)
+			{
+				ActorSystem::tick(m_registry,
+				                  m_registry->get<Player>(*state.first).actor,
```
M3TIOR, line 63: Since you use this twice, you can copy it into a local variable then use that variable twice to save cpu time on reference expansions. It might even make this a slight bit more readable :cheese:.

Phoenix/Server/Source/Iris.cpp
```diff
@@ -70,59 +69,57 @@ Iris::~Iris() { enet_host_destroy(m_server); }
 
 void Iris::run()
 {
-	while (m_running)
+	while (enet_host_service(m_server, &m_event, 1000 / 20) > 0 && m_running)
```
M3TIOR, line 72: If the timeout epoch is always going to be 50 why use a constant funciton? If this is scaffolding for a later epoch adjustment variable it should be commented. If you're dead set on using the current epoch function make it a `constexpr` to make sure all compilers reduce it on compilation rather than run it at runtime, especially because it's in a recursive loop.

Phoenix/Server/Source/Game.cpp
```diff
@@ -0,0 +1,75 @@
...
+	while (m_running)
+	{
+		// @todo @beeper is this efficient?
+		if (!m_iris->stateQueue.front().ready)
+		{
+			std::this_thread::sleep_for(std::chrono::milliseconds(25));
+		}
+		else
+		{
+			printf("State");
```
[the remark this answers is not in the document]
apachano: It's just a debug statement and anything permanent will need to be moved to the logging system instead but that's good to know.

Phoenix/Server/Source/Game.cpp
apachano: The second occurrence is a debug statement again 🤦 I added a comment to make that clearer.

Phoenix/Server/Source/Iris.cpp
apachano: No, it was me being lazy to input 1/20 of a second. Fixed it.

PR #214, opened 2020-04-24
Feat phong shading

SuperFola:
Authors:
@SuperFola 

## Summary of changes
-  adding ambient, diffuse and directional light
- need to send normals and color for each vertex to the shaders
- need to send light color and light position (sun) when rendering world
  
## Caveats
None that I can think about.

## On approval
Please merge

### DIFF
--- a/Phoenix/Assets/SimpleWorld.frag
+++ b/Phoenix/Assets/SimpleWorld.frag
@@ -3,11 +3,11 @@
 in vec3 pass_UV;
 in vec3 pass_normal;
 flat in uint pass_color;
-in vec3 pass_pos;
+// in vec3 pass_pos;
 
 uniform sampler2DArray u_TexArray;
 uniform float u_AmbientStrength;
-uniform vec3 u_LightPos;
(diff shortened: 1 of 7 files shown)


Phoenix/Assets/SimpleWorld.frag
```diff
@@ -1,12 +1,44 @@
...
+uniform float u_Brightness;
 
 out vec4 out_FragColor;
 
 void main()
 {
-	out_FragColor = texture(u_TexArray, pass_UV);
+	int color_i = int(pass_color);
+	vec4 objectColor = vec4(
+		float((color_i & 61440) >> 12) / 16.0,
```
beeperdeeper089, line 19: These are magic numbers (numbers with no clear meaning), please either comment what they are or make the GLSL equivalent of a constexpr variable to name them.

SuperFola: Woops, those numbers are needed to extract 4 bits here and there, to extract the right value of each component

PR #220, opened 2020-05-01
Protocol & States

NicolasRicard:
Adding a Protocol class to add managment of multiple types on same channel
BlockingQueue thread-safe
States queue double queue buffer
NOMINMAX for windows


Phoenix/Server/Include/Server/Server.hpp
```diff
@@ -28,37 +28,34 @@
...
 
-#include <entt/entt.hpp>
 #include <enet/enet.h>
+#include <entt/entt.hpp>
 
-#include <array>
-#include <string>
+#include <Server/Game.hpp>
+#include <Server/Iris.hpp>
+#include <Server/User.hpp>
```
apachano, line 39: Don't flip these around, it should go Server/ on top, Common/ next, third parties third, then std lib last. clang-tidy expects this order to catch duplicate includes.

Phoenix/Server/Source/Game.cpp
```diff
@@ -26,52 +26,25 @@
...
-
-#include <Common/Actor.hpp>
-#include <Common/Movement.hpp>
-#include <Common/Position.hpp>
 #include <thread>
 
-using namespace phx;
-using namespace phx::server;
+#include <Common/Actor.hpp>
+#include <Server/Game.hpp>
```
apachano, line 32: Same thing here, the most local includes first.

Phoenix/Server/Source/Game.cpp
```diff
@@ -26,52 +26,25 @@
...
-		{
-			printf("State");
-			networking::StateBundle m_currentState = m_iris->stateQueue.front();
-			m_iris->stateQueue.pop_front();
-
-			for (auto state : m_currentState.states)
-			{
-				ActorSystem::tick(m_registry,
-				                  m_registry->get<Player>(*state.first).actor,
-				                  dt, state.second);
```
apachano, line 64: Even if the previous system was rudimentary this is a step backward. Could you re-implement a rudimentary system using the new queue?

Phoenix/Server/Source/Iris.cpp
```diff
@@ -26,240 +26,164 @@
...
 #include <Common/Actor.hpp>
-#include <Common/Movement.hpp>
-#include <Common/Position.hpp>
-#include <Common/Serialization/Serializer.hpp>
-
-#include <cstring> //For std::memcpy on non-mac machines
 
-using namespace phx;
-using namespace phx::server::networking;
+#include <Server/Iris.hpp>
```
apachano, line 33: Same as previous

Phoenix/Server/Source/Iris.cpp
```diff
@@ -26,240 +26,164 @@
...
 
 /// @todo Replace this with the config system
 static const std::size_t MAX_USERS = 32;
 
-Iris::Iris(entt::registry* registry, bool* running)
-    : m_registry(registry), m_running(running)
+namespace phx::server::networking
```
apachano, line 38: We probably shouldn't start mixing around the formatting. This adds a ton of changes that make this diff really hard to read since nearly every line has an indentation change and we will need to adjust it back to merge into develop, it just makes everything harder.

Phoenix/Server/Include/Server/Server.hpp
NicolasRicard: I can't change it, I am merging other feat-network

PR #223, opened 2020-05-04
Feat enet cms

apachano:
Authors:
@apachano 

## Summary of changes
-  Add CMS to server
-  Split API between client and server
  
## Caveats
-  No

## On approval
Please merge


Phoenix/Client/Source/Game.cpp
```diff
@@ -83,6 +83,10 @@ Game::Game(gfx::Window* window, entt::registry* registry)
 
 	voxels::BlockRegistry::get()->registerAPI(m_modManager);
 
+	m_modManager->registerFunction(
+	    "core.command.register",
+	    [](std::string command, std::string help, sol::function f) {});
```
SuperFola, line 88: why not use `const std::string&`? same for `sol::function`, would be more cache friendly

Phoenix/Server/Source/Server.cpp
```diff
@@ -42,15 +46,76 @@ Server::Server(std::string save) : m_save(std::move(save))
 	m_game = new Game(&m_registry, &m_running, m_iris);
 }
 
+void registerUnusedAPI(cms::ModManager* manager)
+{
+	manager->registerFunction("core.input.registerInput",
+	                          [](std::string uniqueName,
+	                             std::string displayName,
+	                             std::string defaultKey) {});
```
SuperFola, line 54: same problem here, `const std::string&` would be more cache friendly

apachano: Because constant references to Lua variables starts getting dicey. Last I tried it, it wouldn't work.

SuperFola: strange o.o

Phoenix/Server/Source/Commander.cpp
```diff
@@ -40,68 +40,56 @@
...
-	auto m_command = m_commands.find(command);
-	if (m_command != m_commands.end())
+	if (m_commands.find(command) != m_commands.end())
 	{
-		///@TODO log that we over-wrote a command
-		m_command->second.command  = command;
-		m_command->second.help     = help;
-		m_command->second.callback = f;
-		return;
+		std::cout << "Command overwritten: " << command << "\n";
```
beeperdeeper089, line 92: Replace with logger.
```
LOG_WARNING("MODDING") << "Command overwritten: " << command;
```

PR #231, opened 2020-05-14
Feat enet blocking queue

NicolasRicard:
Add and use BlockingQueue.hpp

### DIFF
--- a/Phoenix/Common/Include/Common/Util/BlockingQueue.hpp
+++ b/Phoenix/Common/Include/Common/Util/BlockingQueue.hpp
@@ -130,24 +130,42 @@ namespace phx
 			return true;
 		}
 
-		void push(T value)
-		{
-			std::lock_guard<std::mutex> lock(m_mutex);
-			if (m_done)
-			{
-				return;
-			}
-			bool unlock = false;
-			if (Container::empty())
-			{
-				unlock = true;
-			}
-			Container::push(value);
-			if (unlock)
-			{
-				m_cond.notify_one();
-			}
-		}
+        void push(const T& value)
+        {
+            std::lock_guard<std::mutex> lock(m_mutex);
+            if (m_done)
+            {
+                return;
+            }
+            bool unlock = false;
+            if (Container::empty())
+            {
+                unlock = true;
+            }
+            Container::push(value);
+            if (unlock)
+            {
+                m_cond.notify_one();
+            }
+        }
+        void push(T&& value)
+        {
+            std::lock_guard<std::mutex> lock(m_mutex);
+            if (m_done)
+            {
+                return;
+            }
+            bool unlock = false;
+            if (Container::empty())
+            {
+                unlock = true;
+            }
+            Container::push(std::move(value));
+            if (unlock)
+            {
+                m_cond.notify_one();
+            }
+        }
 
 		template <class... Args>
 		void emplace(Args&&... args)
--- a/Phoenix/Common/Include/Common/Util/BlockingQueue.hpp
+++ b/Phoenix/Common/Include/Common/Util/BlockingQueue.hpp
@@ -1,0 +1,245 @@
+// Copyright 2019-20 Genten Studios
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice,
+// this list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice,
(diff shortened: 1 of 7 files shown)


Phoenix/Common/Include/Common/Util/BlockingQueue.hpp
```diff
@@ -0,0 +1,245 @@
...
+	{
+	public:
+		virtual ~BlockingQueue() { stop(); }
+
+		void stop()
+		{
+			m_done = true;
+			m_cond.notify_all();
+			m_mutex.lock();
+			m_mutex.unlock();
```
SuperFola, line 62: is there a reason to do so?

Phoenix/Common/Include/Common/Util/BlockingQueue.hpp
```diff
@@ -0,0 +1,245 @@
...
+			if (Container::empty())
+			{
+				return false;
+			}
+			_value = std::move(phx::front<Container>(*this));
+			Container::pop();
+			return true;
+		}
+
+		void push(T value)
```
SuperFola, line 133: how about using a `T&&` and `std::forward` to avoid copies?

Phoenix/Common/Include/Common/Util/BlockingQueue.hpp
```diff
@@ -0,0 +1,245 @@
...
+			if (m_done)
+			{
+				return;
+			}
+			bool unlock = false;
+			if (Container::empty())
+			{
+				unlock = true;
+			}
+			Container::emplace(args...);
```
SuperFola, line 165: `Container::emplace(std::forward(args)...)` would use the correct underlying function (it still works as is but there are numerous unecessary copies)

Phoenix/Common/Include/Common/Util/Front.hpp
```diff
@@ -0,0 +1,71 @@
...
+        template <class Container>
+        struct front
+        {
+            static auto process(Container& t) { return t.front(); }
+        };
+
+        template <>
+        struct front<std::string>
+        {
+            static auto process(std::string& t) { return t.at(1); }
```
SuperFola, line 44: is `.at` wanted? since it's adding a cost with the assertion inside

Phoenix/Server/Source/Iris.cpp
```diff
@@ -161,42 +161,44 @@ void Iris::parseState(entt::entity* userRef, enet_uint8* data, std::size_t dataL
...
 		/// expecting packets from
-		stateQueue.push_back(bundle);
+		currentBundles.push_back(bundle);
 	}
 
 	// Discard state if its older that the oldest stateBundle
-	if (input.sequence < stateQueue.front().sequence &&
-	    stateQueue.back().sequence - input.sequence < 10)
+	if (input.sequence < currentBundles.front().sequence &&
+	    currentBundles.back().sequence - input.sequence < 10)
```
SuperFola, line 176: remove magic number and make it a constexpr

Phoenix/Server/Source/Iris.cpp
```diff
@@ -161,42 +161,44 @@ void Iris::parseState(entt::entity* userRef, enet_uint8* data, std::size_t dataL
...
 		return;
 	}
 
 	// Fill the stateBundles up to the current input sequence
-	while ((input.sequence > stateQueue.back().sequence &&
-	        input.sequence - stateQueue.back().sequence > 10) ||
-	       stateQueue.back().sequence == 255)
+	while ((input.sequence > currentBundles.back().sequence &&
+	        input.sequence - currentBundles.back().sequence > 10) ||
+	       currentBundles.back().sequence == 255)
```
SuperFola, line 185: remove magic number and make it a constexpr

Phoenix/Server/Source/Iris.cpp
```diff
@@ -216,9 +220,11 @@ void Iris::parseState(entt::entity* userRef, enet_uint8* data, std::size_t dataL
 	}
 
 	// If we have more than 10 states enqueued, assume we lost a packet
-	if (stateQueue.size() > 10)
+	if (currentBundles.size() > 10)
```
SuperFola, line 223: same magic number as the one on line 176, same thing

Phoenix/Server/Source/Iris.cpp
NicolasRicard: Well, I didn't change this, so it should not concern the pull request.

Phoenix/Server/Source/Iris.cpp
NicolasRicard: Same, it is not concerned by the pull request

PR #243, opened 2020-06-07
Implement Audio functionality within the Lua implementation

beeperdeeper089:
Resolves: #
Authors:
@beeperdeeper089 

## Summary of changes
-  Fix license in a file
-  Implement lua abstraction for audio system.
-  Play basic background music
-  Implement sounds that will be played on place/break of dirt and grass blocks.
  
## Caveats
-  Audio does not follow the player's position, it will sound as if the player is stuck at 0, 0, 0.

## On approval
Please merge


tobyplowy:
Not too important but would be nice to see that dirt sound also be added to the grass. 
(Already asked in DM but asking in public for the record and all =D )

Phoenix/Client/Include/Client/Audio/Audio.hpp
```diff
@@ -166,6 +172,13 @@ namespace phx::audio
 			return getAudioData(uniqueName);
 		}
 
+		// [] operator, like an unordered_map, just redirected to getAudioData,
+		// essentially just a QoL feature.
+		AudioData operator[](Handle handle) const
```
SuperFola, line 177: AudioData, not const AudioData& ?

Phoenix/Client/Source/Audio/Source.cpp
```diff
@@ -36,81 +36,77 @@ using namespace phx::audio;
...
 
 Source::Source(AudioData data)
 {
+	m_source = {new unsigned int {}, Deleter()};
+
 	// generates an OpenAL source and sets the buffer to be used during
 	// playback.
-	alGenSources(1, &m_source);
-	alSourcei(m_source, AL_BUFFER, data.buffer);
+	alGenSources(1, &*m_source);
```
SuperFola, line 51: &*var isn't the same as just var?

beeperdeeper089: M_source is a shared pointer, you have to dereference to get the value and then get the address of the underlying variable (the actual unsigned int rather than the container).

SuperFola: oh i missed the part where it was defined as a shared ptr then, sorry ^^

Phoenix/Client/Include/Client/Audio/Audio.hpp
beeperdeeper089: We have to return an empty copy when the handle/uniquename doesn't exist so returning a literal copy makes it easier.

PR #246, opened 2020-06-12
Feat enet abstraction

apachano:
## Summary of changes
-  Implements netcode abstraction into existing networking code
  
## Caveats
-  Terminal has a flushing issue is prints one message behind

## On approval
Please merge

### DIFF
--- a/Phoenix/Client/Include/Client/Game.hpp
+++ b/Phoenix/Client/Include/Client/Game.hpp
@@ -95,13 +95,13 @@ namespace phx::client
 
 		cms::ModManager* m_modManager;
 
-		Crosshair* m_crosshair = nullptr;
+		Crosshair*  m_crosshair  = nullptr;
 		EscapeMenu* m_escapeMenu = nullptr;
 		GameTools*  m_gameDebug  = nullptr;
 		bool        m_followCam  = true;
 		math::vec3  m_prevPos;
 		int         m_playerHand = 0;
 
-		client::net::Network* m_network;
+		client::Network* m_network;
 	};
 } // namespace phx::client
--- a/Phoenix/Client/Include/Client/Game.hpp
+++ b/Phoenix/Client/Include/Client/Game.hpp
@@ -78,7 +78,7 @@ namespace phx::client
 		 * @param cout Needs to be depreciated, unused (but required by
 		 * terminal)
 		 */
-		void sendMessage(const std::string& input, std::ostringstream& cout);
+		void sendMessage(std::string input, std::ostringstream& cout);
 		void parseEvent(enet_uint8* data, std::size_t dataLength);
 		void parseState(enet_uint8* data, std::size_t dataLength);
 		void parseMessage(enet_uint8* data, std::size_t dataLength);
--- a/Phoenix/Client/Include/Client/Game.hpp
+++ b/Phoenix/Client/Include/Client/Game.hpp
@@ -28,6 +28,9 @@
 
 #pragma once
 
+#include <Client/Crosshair.hpp>
+#include <Client/EscapeMenu.hpp>
+#include <Client/GameTools.hpp>
 #include <Client/Graphics/Camera.hpp>
 #include <Client/Graphics/Layer.hpp>
 #include <Client/Graphics/ShaderPipeline.hpp>
@@ -35,8 +38,6 @@
 #include <Client/Graphics/Window.hpp>
 #include <Client/Network.hpp>
 #include <Client/Player.hpp>
-#include <Client/Crosshair.hpp>#include <Client/EscapeMenu.hpp>
-#include <Client/GameTools.hpp>
 
 #include <Common/CMS/ModManager.hpp>
 
--- a/Phoenix/Client/Include/Client/Game.hpp
+++ b/Phoenix/Client/Include/Client/Game.hpp
@@ -28,8 +28,6 @@
 
 #pragma once
 
-#include <Client/EscapeMenu.hpp>
-#include <Client/GameTools.hpp>
 #include <Client/Graphics/Camera.hpp>
 #include <Client/Graphics/Layer.hpp>
 #include <Client/Graphics/ShaderPipeline.hpp>
@@ -37,6 +35,8 @@
 #include <Client/Graphics/Window.hpp>
 #include <Client/Network.hpp>
 #include <Client/Player.hpp>
+#include <Client/Crosshair.hpp>#include <Client/EscapeMenu.hpp>
+#include <Client/GameTools.hpp>
 
 #include <Common/CMS/ModManager.hpp>
 
@@ -94,6 +94,7 @@ namespace phx::client
 
 		cms::ModManager* m_modManager;
 
+		Crosshair* m_crosshair = nullptr;
 		EscapeMenu* m_escapeMenu = nullptr;
 		GameTools*  m_gameDebug  = nullptr;
 		bool        m_followCam  = true;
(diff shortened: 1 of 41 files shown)


beeperdeeper089:
@apachano I'm going to merge develop into feat-enet if I can do so cleanly and hopefully this PR's size reduces a lot lol

Phoenix/Client/Include/Client/Game.hpp
```diff
@@ -79,10 +80,7 @@ namespace phx::client
 		 * @param cout Needs to be depreciated, unused (but required by
 		 * terminal)
 		 */
-		void sendMessage(const std::string& input, std::ostringstream& cout);
-		void parseEvent(enet_uint8* data, std::size_t dataLength);
-		void parseState(enet_uint8* data, std::size_t dataLength);
-		void parseMessage(enet_uint8* data, std::size_t dataLength);
+		void sendMessage(std::string input, std::ostringstream& cout);
```
beeperdeeper089, line 83: const std::string&

Phoenix/Client/Include/Client/Network.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+		void kill() { m_running = false; };
+
+		/**
+		 * @brief Actions taken when a state is received
+		 *
+		 * @param data The data in the state packet
+		 * @param dataLength The length of the data in the state packet
+		 */
+		void parseEvent(phx::net::Packet& packet);
```
beeperdeeper089, line 53: You don't need to put the ``phx`` here, this mainly just applies to source files.

Phoenix/Client/Include/Client/Network.hpp
```diff
@@ -0,0 +1,92 @@
...
+		 */
+		void parseMessage(phx::net::Packet& packet);
+
+		/**
+		 * @brief Sends a state packet to a client
+		 *
+		 * @param userRef The user to sent the state to
+		 * @param data The state packet data
+		 */
+		void sendState(InputState inputState);
```
beeperdeeper089, line 77: const InputState&

Phoenix/Client/Include/Client/Network.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+		/**
+		 * @brief Sends a message packet to a client
+		 *
+		 * @param userRef The user to sent the message to
+		 * @param data The message packet data
+		 */
+		void sendMessage(std::string message);
```
beeperdeeper089, line 85: const std::string&

Phoenix/Client/Source/Game.cpp
```diff
@@ -365,33 +277,27 @@ void Game::tick(float dt)
...
+		inputState.sequence = sequence;
+		inputState.forward  = m_window->isKeyDown(events::Keys::KEY_W);
+		inputState.backward = m_window->isKeyDown(events::Keys::KEY_S);
+		inputState.left     = m_window->isKeyDown(events::Keys::KEY_A);
+		inputState.right    = m_window->isKeyDown(events::Keys::KEY_D);
+		inputState.up       = m_window->isKeyDown(events::Keys::KEY_SPACE);
+		inputState.down     = m_window->isKeyDown(events::Keys::KEY_LEFT_SHIFT);
+
+		/// conversion from rad to 1/1000 of degrees
+		inputState.rotation.x = static_cast<unsigned>(
```
beeperdeeper089, line 292: use unsigned int instead of unsigned, just for clarity (won't make the slightest difference but just a consideration)

Phoenix/Client/Include/Client/Network.hpp
```diff
@@ -0,0 +1,92 @@
...
+#include <Common/Input.hpp>
+#include <Common/Network/Host.hpp>
+#include <Common/Util/BlockingQueue.hpp>
+
+namespace phx::client
+{
+	class Network
+	{
+	public:
+		Network(std::ostringstream& chat);
```
beeperdeeper089, line 40: I think we should try not to couple the network with the chat, there must be another way to do this reliably. Potentially an onEnter callback or something on the chat class as well as a callback to push things to the chat in the Network class itself?

This will suffice for now but I personally think it would be better to decouple the two systems.

Phoenix/Client/Source/Game.cpp
```diff
@@ -365,33 +277,27 @@ void Game::tick(float dt)
...
+		    m_registry->get<Position>(m_player->getEntity()).rotation.x *
+		    360000.0);
```
beeperdeeper089, line 294: Magic number, use a ``constexpr static float`` and just name it something that makes more sense than just ``360000.0``

Phoenix/Client/Source/Game.cpp
```diff
@@ -365,33 +277,27 @@ void Game::tick(float dt)
...
+		inputState.rotation.y = static_cast<unsigned>(
+		    m_registry->get<Position>(m_player->getEntity()).rotation.y *
+		    360000.0);
```
beeperdeeper089, line 297: Same as above.

Phoenix/Common/Source/Network/Host.cpp
```diff
@@ -204,12 +213,17 @@ void Host::handleEvent(ENetEvent& event)
...
+		return m_peers.at(std::size_t(peer.data));
+	}
+	std::cout << "peer not found";
 }
 
 Peer& Host::createPeer(ENetPeer& peer)
 {
-	peer.data    = reinterpret_cast<void*>(m_peerID++);
+	m_peerID++;
+	peer.data = reinterpret_cast<void*>(m_peerID);
```
beeperdeeper089, line 226: Use ++m_peerID (compiler will optimise it out but using the prefix incremental operator, you don't make an unnecessary copy). You could then probably consolidate the lines to:
```cpp
peer.data = reinterpret_cast<void*>(++m_peerID)
```

Phoenix/Server/Source/Iris.cpp
```diff
@@ -194,24 +156,24 @@ void Iris::parseState(entt::entity* userRef, enet_uint8* data, std::size_t dataL
 	}
 
 	{
-		//		printf("insert existing %lu \n", input.sequence);
+		// printf("insert existing %lu \n", input.sequence);
 		auto begin = currentBundles.begin();
```
beeperdeeper089, line 160: Is this not a redundant declaration?

Phoenix/Client/Include/Client/Network.hpp
apachano: Earlier I had a client::net and it couldn't differentiate between the two. Removed it since it was redundant anyways

Phoenix/Client/Include/Client/Network.hpp
apachano: I agree, I'm not a fan of having the networking system coupled at all but this was the simpler of my options to get the text to the terminal. Its worth noting that while I named the argument "chat" its actually just the address of a stream and cout could be substituted so it's not really coupled to the chat, it just needs a place to dump output. 

Regardless, definitely something that needs cleaned up in the future.

Phoenix/Client/Include/Client/Network.hpp
apachano: image

Phoenix/Client/Include/Client/Game.hpp
apachano: image

Phoenix/Client/Include/Client/Network.hpp
beeperdeeper089: Ah, ok this is kinda funky, leave it like this then, I know a fix that will work and I'll look to getting that through soon.

Issue #43, opened 2020-01-07
Collision


apachano:
## Story
The player should collide with any solid object type

## MVP
What are the minimum features needed for this story?
[ ] Player cant move through solid objects

## Stretch
What are some non-required features for this story?
[ ] Player has setting to disable collision

tobyplowy:
"Player has setting to disable collision" This will come in use for modding tho it's not really a 0.1 feature but I don't think it would be too hard to add I think right?

apachano:
It should be extremely easy to add, it will probably be something that you can toggle via a command but it can easily be exposed to modding as well. I left it as a stretch because it's not really mandatory.

tobyplowy:
@apachano Yeah :)

beeperdeeper089:
Oof, this is going to be a difficult task, we're going to have to research on it.

PR #250, opened 2020-06-22
adding first basic collision system, with a very slow gravity

SuperFola:
Resolves: #43 
Authors:
@SuperFola 

## Summary of changes
- adding a `setWorld` (temporary, but since it's the camera who's handling the movement atm) to the camera
- updating the camera.tick function to add collision detection and gravity (very basic, should be updated in the future)
  
## Caveats
- falling into the void if the player spawns with no block below them
- falling while flying

## On approval
Wait for me for confirmation.


beeperdeeper089:
I personally think rather than hacking this in, we should probably implement the movement system (remember that we're going to try use an ECS design) and have that do it. This will make the eventual integration of the server less painful and will reduce the coupling of our code at this point in time.

If we keep this (it's fine for now if we decide it's not worth the effort), we NEED to implement a flying mode otherwise if we're testing something and we accidently fall off the world then we just fall forever, and we also need to change the spawn position from 0, 0, 0 to the top or something where we don't just fall and die to start with.

tobyplowy:
I have been testing this with tears of joy in my eyes. 
This is amazing! Thanks Fola. 

Now for some of the issues I found while testing:
•While holding shift and standing on the ground the camera sinks underground? 
•You can walk into walls.
•The camera should be 2 blocks tall instead of the current 1 block tall. 
And the above stated by Vyom.

tobyplowy:
@SuperFola Can you fix the conflicts and add wall collision so this PR can be merged?

beeperdeeper089:
@SuperFola if we get the conflicts resolved on here, i'll get this reviewed asap and we can try to get this stuff in.

tobyplowy:
Closing due to inactivity. Please feel free to request a reopening if you do decide to continue work.

PR #255, opened 2020-06-25
Implement logging into Lua

apachano:
Authors:
@apachano 

## Summary of changes
- Implements Logging API in Lua
  
## Caveats
- API can't be implemented in Logging.hpp else a circular dependency is formed and breaks compilation
- There is no way to reliably determine which mod calls the logging function so all modules log as the same component

## On approval
Merge

### DIFF
--- a/Phoenix/Client/Source/Game.cpp
+++ b/Phoenix/Client/Source/Game.cpp
@@ -98,6 +98,19 @@ Game::Game(gfx::Window* window, entt::registry* registry)
 	InputMap::get()->registerAPI(m_modManager);
 	CommandBook::get()->registerAPI(m_modManager);
 
+	m_modManager->registerFunction("core.log_warning", [](std::string message) {
+		LOG_WARNING("MODULE") << message;
+	});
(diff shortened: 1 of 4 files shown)


Phoenix/Client/Source/Game.cpp
```diff
@@ -98,6 +98,19 @@ Game::Game(gfx::Window* window, entt::registry* registry)
 	InputMap::get()->registerAPI(m_modManager);
 	CommandBook::get()->registerAPI(m_modManager);
 
+	m_modManager->registerFunction("core.log_warning", [](std::string message) {
+		LOG_WARNING("MODULE") << message;
```
beeperdeeper089, line 102: Is there anything we could use instead of module? It just doesn't feel like the right word

apachano: I'm up for suggestions. Ideally, the module name would go there but like I said in the caveats, that is not easy to reliably obtain.

apachano: Gonna merge anyways, this is an easy change if anyone comes up with something better or figured out actually including the module name.

PR #258, opened 2020-07-03
Feat enet predictions

apachano:
Authors:
@apachano 

## Summary of changes
- Implements prediction confirmation and resets

## On approval
Merge

### DIFF
--- a/Phoenix/Client/Include/Client/Graphics/Camera.hpp
+++ b/Phoenix/Client/Include/Client/Graphics/Camera.hpp
@@ -221,7 +221,8 @@ namespace phx::gfx
 
 		math::vec2 m_windowCentre;
 
-		bool m_enabled = true;
+		// Also used by InputQueue to determine if camera is focused or not
+		std::atomic<bool> m_enabled = true;
 
 		Setting* m_settingSensitivity;
 
--- a/Phoenix/Client/Include/Client/Graphics/Camera.hpp
+++ b/Phoenix/Client/Include/Client/Graphics/Camera.hpp
@@ -215,11 +215,7 @@ namespace phx::gfx
 		Window* m_window;
 
 		math::mat4 m_projection;
-
-		math::vec3 m_rotation;
-		math::vec3 m_position;
 		math::vec3 m_up;
-		math::vec3 m_direction;
 
 		math::vec2 m_windowCentre;
 
@@ -230,14 +226,6 @@ namespace phx::gfx
 
 		entt::registry* m_registry;
 		entt::entity    m_actor;
-
-		/// @todo Relocate player movement to player instead of camera
-		client::Input* m_forward;
-		client::Input* m_backward;
-		client::Input* m_left;
-		client::Input* m_right;
-		client::Input* m_fly;
-		client::Input* m_down;
 	};
 } // namespace phx::gfx
 
--- a/Phoenix/Client/Include/Client/Graphics/Camera.hpp
+++ b/Phoenix/Client/Include/Client/Graphics/Camera.hpp
@@ -194,7 +194,7 @@ namespace phx::gfx
 		math::vec2 m_windowCentre;
 
 		// Also used by InputQueue to determine if camera is focused or not
-		std::atomic<bool> m_enabled = true;
+		std::atomic<bool> m_enabled;
 
 		Setting* m_settingSensitivity;
 
--- a/Phoenix/Client/Include/Client/Graphics/Camera.hpp
+++ b/Phoenix/Client/Include/Client/Graphics/Camera.hpp
@@ -44,6 +44,8 @@
 
 #include <entt/entt.hpp>
 
+#include <atomic>
+
 namespace phx::gfx
 {
 	/**
--- a/Phoenix/Client/Include/Client/Graphics/Camera.hpp
+++ b/Phoenix/Client/Include/Client/Graphics/Camera.hpp
@@ -92,32 +92,6 @@ namespace phx::gfx
 		explicit FPSCamera(Window* window, entt::registry* registry);
 
 		/**
-		 * @brief Gets the position of the camera.
-		 * @return The 3-component world-space position.
-		 *
-		 * This camera return is NOT compatible with the voxel world. The
-		 * coordinate system used by the camera and the system used by the
-		 * voxel world are different, because of the way block coordinates
-		 * have been setup.
-		 */
-		math::vec3 getPosition() const;
-
-		/**
-		 * @brief Gets the direction the camera is facing.
-		 * @return The 3-component vector representing direction.
-		 *
-		 * The direction is always the direction the player is looking, this
-		 * can be used with the Ray object to find what the player is
-		 * actually looking at within the actual world, whether it be voxels
-		 * or something else. Because this is not representative of
-		 * position, and direction cannot be skewed by the world coordinate
-		 * system, this data is compatible with every other layer requiring
-		 * and supporting direction, of course when asking for a 3
-		 * dimensional vector.
-		 */
-		math::vec3 getDirection() const;
-
-		/**
 		 * @brief Sets the camera perspective projection.
 		 * @param projection The actual projection to set the camera to.
 		 *
--- a/Phoenix/Client/Source/Game.cpp
+++ b/Phoenix/Client/Source/Game.cpp
@@ -445,7 +445,7 @@ void Game::sendMessage(const std::string& input, std::ostringstream& cout)
 	m_network->sendMessage(input);
 }
 
-void Game::confirmState(const Position& position)
+void Game::confirmState(Position position)
 {
 	static std::list<InputState> states;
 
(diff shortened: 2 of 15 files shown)


Phoenix/Client/Include/Client/Graphics/Camera.hpp
```diff
@@ -213,28 +189,17 @@ namespace phx::gfx
...
-		math::vec3 m_rotation;
-		math::vec3 m_position;
 		math::vec3 m_up;
-		math::vec3 m_direction;
 
 		math::vec2 m_windowCentre;
 
-		bool m_enabled = true;
+		// Also used by InputQueue to determine if camera is focused or not
+		std::atomic<bool> m_enabled = true;
```
SuperFola, line 197: just wondering but maybe it would be better to have this initialized in the constructor?

Phoenix/Client/Source/Game.cpp
```diff
@@ -455,3 +460,57 @@ void Game::sendMessage(const std::string& input, std::ostringstream& cout)
 {
 	m_network->sendMessage(input);
 }
+
+void Game::confirmState(const Position& position)
+{
+	// This is an internal log of the recent input states sent to the server
+	static std::list<InputState> states;
```
beeperdeeper089, line 467: I would recommend against making this static.

Phoenix/Client/Source/Game.cpp
```diff
@@ -455,3 +460,57 @@ void Game::sendMessage(const std::string& input, std::ostringstream& cout)
...
+	{
+		phx::ActorSystem::tick(m_registry, entity, 1.f / 20.f, inputState);
+	}
+
+	// Compare the temporary position object with the current player position.
+	math::vec3 diff = position.position - pos.position;
+
+	// If the temporary position object based on the confirmation is too far off
+	// from our current position, update our current position.
+	const float precision = .25f;
```
beeperdeeper089, line 507: The precision should probably be the distance you could theoretically travel in a few ticks, let's say 10. If that makes sense?

Phoenix/Client/Source/Network.cpp
```diff
@@ -94,7 +94,28 @@ void Network::parseEvent(phx::net::Packet& packet)
 	std::cout << "Event received";
 }
 
-void Network::parseState(phx::net::Packet& packet) {}
+void Network::parseState(phx::net::Packet& packet)
+{
+	static size_t currentSequence;
```
beeperdeeper089, line 99: Again, I would recommend against a static variable for this use case.

Phoenix/Common/Include/Common/Input.hpp
```diff
@@ -36,14 +36,14 @@ namespace phx
...
-		bool        down{};
-		math::vec2u rotation{}; // in 1/1000 degres
-		std::size_t sequence{};
+		bool        forward {};
+		bool        backward {};
+		bool        left {};
+		bool        right {};
+		bool        up {};
+		bool        down {};
+		math::vec2i rotation {}; // in 1/1000 degres
```
beeperdeeper089, line 45: Should rotation not be a 3 dimensional property?

Phoenix/Server/Source/Game.cpp
```diff
@@ -53,20 +51,19 @@ void Game::run()
 	while (m_running)
 	{
 		// Process everybody's input first
+		if (m_iris->stateQueue.empty())
```
beeperdeeper089, line 54: You might be able to use a condition_variable to reduce wasted CPU time further, however it may not be suitable so take this with a pinch of salt.

Phoenix/Client/Include/Client/Graphics/Camera.hpp
apachano: Moved

Phoenix/Client/Source/Game.cpp
apachano: If it's not static it needs to become a member variable instead so it persists. I haven't moved it yet because I am completely lost as to where, but honestly, this should live somewhere besides the game object itself. 

I'd definitely love your thoughts here, personally I wish we didn't have the buffer but the blocking queue is a queue and we need a list to iterate through.

Phoenix/Client/Source/Game.cpp
apachano: Yes and No, it really depends on how strong your CPU and network connection are so this should probably move to a config file. On a strong computer and a local network we rarely if ever need to reset even if the server is off by 10. When it does reset, it's a huge jump. 

For my machine, .25 didn't kill the performance but the updates were frequent and minimal so I barely noticed them. At 1 the updates weren't very frequent and when they did hit it was a fairly annoying jolt.

Phoenix/Client/Source/Network.cpp
apachano: Moved to a member variable

Phoenix/Common/Include/Common/Input.hpp
apachano: Fixed

Phoenix/Server/Source/Game.cpp
apachano: That would make sense, but either the blocking queue itself needs to be changed to support this, or we need to track another piece of data which increases complexity and the potential to run into a bug. I am all for it, but the build may be costly for the performance increase we see.

Phoenix/Client/Source/Game.cpp
beeperdeeper089: I personally would make it a member variable of the game then, keeping it static doesn't make much sense.

Phoenix/Client/Source/Game.cpp
```diff
@@ -455,3 +460,57 @@ void Game::sendMessage(const std::string& input, std::ostringstream& cout)
...
 	m_network->sendMessage(input);
 }
+
+void Game::confirmState(const Position& position)
+{
+	// This is an internal log of the recent input states sent to the server
+	static std::list<InputState> states;
+
+	// We can't iterate through a queue so we drain the queue to this list
+	size_t i = m_inputQueue->m_queue.size();
```
beeperdeeper089, line 470: While using ``size_t`` is just fine, I would recommend you use ``std::size_t`` just so we're sticking to c++ rather than dipping into c.

Phoenix/Client/Source/Game.cpp
```diff
@@ -455,3 +460,57 @@ void Game::sendMessage(const std::string& input, std::ostringstream& cout)
...
+	for (size_t j = 0; j <= i; j++)
```
beeperdeeper089, line 471: Same ``size_t`` concern here.

Phoenix/Client/Source/Graphics/Camera.cpp
```diff
@@ -102,61 +82,19 @@ void FPSCamera::tick(float dt)
 
 	m_window->setCursorPosition(m_windowCentre);
 
-	const float sensitivity = static_cast<float>(m_settingSensitivity->value()) / 50;
+	const float sensitivity =
+	    static_cast<float>(m_settingSensitivity->value()) / 50;
```
beeperdeeper089, line 86: Use ``50.f``, probably won't make a difference but let's be consistent if we want an actual floating point division.

Phoenix/Common/Include/Common/Position.hpp
```diff
@@ -41,12 +41,35 @@ namespace phx
...
+		math::vec3 getDirection() const
 		{
 			return math::vec3 {std::cos(rotation.y) * std::sin(rotation.x),
 			                   std::sin(rotation.y),
 			                   std::cos(rotation.y) * std::cos(rotation.x)};
 		};
+
+		math::vec3 getRight() const
+		{
+			return math::vec3 {std::sin(rotation.x - math::PIDIV2), 0.f,
```
beeperdeeper089, line 66: Do we want to force ``0.f`` for the y-axis here? If we do, it will be impossible to implement any "roll".

Phoenix/Common/Include/Common/Position.hpp
```diff
@@ -41,12 +41,35 @@ namespace phx
...
+			                   std::cos(rotation.x - math::PIDIV2)};
+		};
+
+		math::vec3 getForward() const
+		{
+			return math::vec3 {std::sin(rotation.x), 0.f, std::cos(rotation.x)};
```
beeperdeeper089, line 72: Same concern with potential implementation of roll.

Phoenix/Server/Source/Game.cpp
beeperdeeper089: I'll take a look at the BlockingQueue and see if I can update it to meet our requirements, it should be fine for now though.

Phoenix/Common/Include/Common/Position.hpp
apachano: Honestly, I just copied the existing math to avoid letting things get messy. If we do want to implement roll later then we can actually return something here or add a second movement system entirely.

Phoenix/Common/Include/Common/Position.hpp
apachano: Same

apachano:
@beeperdeeper089 Requesting re-review

Issue #111, opened 2020-02-06
Having the possibility of using texture resolutions other then X16


tobyplowy:
## Story
As of right now it's impossible to use textures other than perfect 16X16 this prevents the use of textures of any other resolution.

## MVP
What are the minimum features needed for this story?
-  [ ] Allow for any power of 2 resolution of texture resolutions. (As far as we can allow)

## Stretch
What are some non-required features for this story?
-  [ ] Support resolutions that are not power of 2.

tobyplowy:
May need to be added to make #132 work better.

PR #261, opened 2020-07-07
WIP: Revised Texture Packer

beeperdeeper089:
Resolves: #111
Authors:
@beeperdeeper089 

## Summary of changes
- Implement Texture Packer
    - Loads ~8 million 16px textures, and upto ~8k 512px textures, and everything in between is between those numbers lol, much more than most people should ever need. Most people will run out of VRAM before they can even fit in 8,000 512px textures (~8GB of VRAM).
  
## Caveats
- Only Power of 2 textures can be used for blocks/items.
- If a GPU memory allocation fails, the application just aborts (that's what happens when there's a bad malloc/new anyway so not that bad.)
- No mipmapping or filtering as of right now, future implementation is planned and not too difficult.

## On approval
Wait for me, I have some stuff to get into the repo before this can go in.


tobyplowy:
Resolves https://github.com/GentenStudios/Phoenix/issues/111

Issue #149, opened 2020-03-16
Send voxel world to client


apachano:
## Story
When a client connects, it should get the state of the chunks around the player, when the player moves the client should get new chunks.

## MVP
What are the minimum features needed for this story?
- [ ] On first connection, n chunks around the player are transmitted to the client.
- [x] When the player moves into a new chunk, new chunks are sent to the client

tobyplowy:
@apachano I think it would be much better if you only send relevant data and not the entire word?
So only the chunks/metadata that the player needs at that time? That way you are not sending and receiving gigs of data in one go. (Just thinking out loud)

beeperdeeper089:
I second that, let's only send what needs to be sent.

From personal experience a server with 20-30 people exploring the world can result in a shit load of chunks which is a lot of data

apachano:
Updated, IDK what I was thinking when I typed this. Maybe MVP?

PR #263, opened 2020-07-10
Feat enet voxels

apachano:
Resolves: #149 
Authors:
@apachano 

## Summary of changes
- Transmits chunks to the client as they move around
- Hides client networking behind a flag so local play is still possible.

## On approval
Merge

### DIFF
--- a/Phoenix/Client/Include/Client/Game.hpp
+++ b/Phoenix/Client/Include/Client/Game.hpp
@@ -62,7 +62,8 @@ namespace phx::client
 	class Game : public gfx::Layer
 	{
 	public:
-		explicit Game(gfx::Window* window, entt::registry* registry);
+		explicit Game(gfx::Window* window, entt::registry* registry,
+		              bool networked = true);
 		~Game() override;
 
 		void onAttach() override;
@@ -112,8 +113,8 @@ namespace phx::client
 		math::vec3  m_prevPos;
 		int         m_playerHand = 0;
 
-		client::Network*    m_network;
-		client::InputQueue* m_inputQueue;
+		client::Network*    m_network    = nullptr;
+		client::InputQueue* m_inputQueue = nullptr;
 
 		// intermediary variables to prevent getting the pointer from the client
 		// singleton every tick.
--- a/Phoenix/Client/Include/Client/Game.hpp
+++ b/Phoenix/Client/Include/Client/Game.hpp
@@ -115,6 +115,8 @@ namespace phx::client
 
 		client::Network*    m_network    = nullptr;
 		client::InputQueue* m_inputQueue = nullptr;
+		// This is an internal log of the recent input states sent to the server
+		std::list<InputState> m_states;
 
 		// intermediary variables to prevent getting the pointer from the client
 		// singleton every tick.
--- a/Phoenix/Client/Include/Client/Network.hpp
+++ b/Phoenix/Client/Include/Client/Network.hpp
@@ -40,7 +40,7 @@ namespace phx::client
 	class Network
 	{
 	public:
-		Network(std::ostringstream& chat, phx::net::Address address);
+		Network(std::ostringstream& chat, const phx::net::Address& address);
 		~Network();
 
 	private:
--- a/Phoenix/Client/Include/Client/Network.hpp
+++ b/Phoenix/Client/Include/Client/Network.hpp
@@ -32,6 +32,7 @@
 #include <Common/Network/Host.hpp>
 #include <Common/Position.hpp>
 #include <Common/Util/BlockingQueue.hpp>
+#include <Common/Voxels/Chunk.hpp>
 #include <thread>
 
 namespace phx::client
@@ -74,6 +75,8 @@ namespace phx::client
 		 */
 		void parseMessage(phx::net::Packet& packet);
 
+		void parseData(phx::net::Packet& packet);
+
 	public:
 		/**
 		 * @brief Sends a state packet to a client
@@ -92,9 +95,10 @@ namespace phx::client
 		void sendMessage(std::string message);
 
 		phx::BlockingQueue<std::pair<Position, size_t>> stateQueue;
+		phx::BlockingQueue<voxels::Chunk>               chunkQueue;
 
 	private:
-		bool                m_running;
+		bool                m_running = false;
 		phx::net::Host*     m_client;
 		std::ostringstream& m_chat;
 		std::thread         m_thread;
--- a/Phoenix/Client/Include/Client/Network.hpp
+++ b/Phoenix/Client/Include/Client/Network.hpp
@@ -40,7 +40,7 @@ namespace phx::client
 	class Network
 	{
 	public:
-		Network(std::ostringstream& chat);
+		Network(std::ostringstream& chat, phx::net::Address address);
 		~Network();
 
 	private:
--- a/Phoenix/Client/Source/Game.cpp
+++ b/Phoenix/Client/Source/Game.cpp
@@ -486,14 +486,11 @@ void Game::sendMessage(const std::string& input, std::ostringstream& cout)
 
 void Game::confirmState(const Position& position)
 {
-	// This is an internal log of the recent input states sent to the server
-	static std::list<InputState> states;
-
 	// We can't iterate through a queue so we drain the queue to this list
-	size_t i = m_inputQueue->m_queue.size();
-	for (size_t j = 0; j <= i; j++)
+	std::size_t i = m_inputQueue->m_queue.size();
+	for (std::size_t j = 0; j <= i; j++)
 	{
-		states.push_back(m_inputQueue->m_queue.pop());
+		m_states.push_back(m_inputQueue->m_queue.pop());
 	}
 
 	// If there are no confirmation states ready from the network, we are done
@@ -504,9 +501,9 @@ void Game::confirmState(const Position& position)
 	auto confirmation = m_network->stateQueue.pop();
 
 	// Discard any inputStates older than the confirmationState
-	while (!states.empty() && states.front().sequence < confirmation.second)
+	while (!m_states.empty() && m_states.front().sequence < confirmation.second)
 	{
-		states.pop_front();
+		m_states.pop_front();
 	}
 
 	// Create a temporary position entity located at the confirmation position
@@ -517,7 +514,7 @@ void Game::confirmState(const Position& position)
         entity, confirmation.first.rotation, confirmation.first.position);
 	m_registry->emplace<Movement>(
 	    entity, m_registry->get<Movement>(m_player->getEntity()).moveSpeed);
-	for (const auto& inputState : states)
+	for (const auto& inputState : m_states)
 	{
 		phx::ActorSystem::tick(m_registry, entity, 1.f / 20.f, inputState);
 	}
--- a/Phoenix/Client/Source/Game.cpp
+++ b/Phoenix/Client/Source/Game.cpp
@@ -37,8 +37,8 @@
 #include <Common/Commander.hpp>
 #include <Common/Position.hpp>
 #include <Common/Logger.hpp>
-
 #include <Common/Movement.hpp>
+
 #include <cmath>
 #include <tuple>
 
--- a/Phoenix/Client/Source/Game.cpp
+++ b/Phoenix/Client/Source/Game.cpp
@@ -59,10 +59,24 @@ static void rawEcho(const std::string& input, std::ostringstream& cout)
 	myGame->sendMessage(input, cout);
 }
 
-Game::Game(gfx::Window* window, entt::registry* registry)
+Game::Game(gfx::Window* window, entt::registry* registry, bool networked)
     : Layer("Game"), m_window(window), m_registry(registry)
 {
+	m_chat = new ui::ChatWindow("Chat Window", 5,
+	                            "Type /help for a command list and help.");
+
+	/// @TODO replace with network callback
+	m_chat->registerCallback(rawEcho);
+
+	if (networked)
+	{
+		m_network = new client::Network(m_chat->cout,
+		                                phx::net::Address("127.0.0.1", 7777));
+	}
+	// else TODO enable this else when we get mod list from network
+	//{
 	const std::string save = "save1";
+	//}
 
 	std::fstream             fileStream;
 	std::vector<std::string> toLoad;
@@ -253,14 +267,10 @@ Game::~Game() { delete m_chat; }
 
 void Game::onAttach()
 {
-	m_chat = new ui::ChatWindow("Chat Window", 5,
-	                            "Type /help for a command list and help.");
-
-	/// @TODO replace with network callback
-	m_chat->registerCallback(rawEcho);
-
-	m_network = new client::Network(m_chat->cout);
-	m_network->start();
+	if (m_network != nullptr)
+	{
+		m_network->start();
+	}
 
 	m_player = new Player(m_registry);
 	m_player->registerAPI(m_modManager);
@@ -276,7 +286,14 @@ void Game::onAttach()
 
 	LOG_INFO("MAIN") << "Registering world";
 	const std::string save = "save1";
-	m_world = new voxels::ChunkView(3, voxels::Map(save, "map1"));
+	if (m_network != nullptr)
+	{
+		m_world = new voxels::ChunkView(3, m_network);
+	}
+	else
+	{
+		m_world = new voxels::ChunkView(3, new voxels::Map(save, "map1"));
+	}
 	m_player->setWorld(m_world);
 	m_camera = new gfx::FPSCamera(m_window, m_registry);
 	m_camera->setActor(m_player->getEntity());
@@ -308,18 +325,21 @@ void Game::onAttach()
 	}
 
 	m_inputQueue = new InputQueue(m_registry, m_player, m_camera);
-	m_inputQueue->start(std::chrono::milliseconds(50), m_network);
+	if (m_network != nullptr)
+	{
+		m_inputQueue->start(std::chrono::milliseconds(50), m_network);
+	}
 
 	LOG_INFO("MAIN") << "Game layer attached";
 }
 
 void Game::onDetach()
 {
-	m_network->stop();
 	delete m_world;
+	delete m_inputQueue;
+	delete m_network;
 	delete m_player;
 	delete m_camera;
-	delete m_network;
 }
 
 void Game::onEvent(events::Event& e)
@@ -430,7 +450,10 @@ void Game::tick(float dt)
 	InputState state = m_inputQueue->getCurrentState();
 
 	ActorSystem::tick(m_registry, m_player->getEntity(), dt, state);
-	confirmState(position);
+	if (m_network != nullptr)
+	{
+		confirmState(position);
+	}
 
 	if (m_followCam)
 	{
--- a/Phoenix/Client/Source/Graphics/ChunkView.cpp
+++ b/Phoenix/Client/Source/Graphics/ChunkView.cpp
@@ -76,22 +76,14 @@ void ChunkView::tick(math::vec3 playerPos)
 	}
 
 	// this converts the raw camera/player position into voxel-world positions.
-	playerPos = playerPos / 2.f;
-	playerPos += 0.5f;
+	playerPos = (playerPos / 2.f) + 0.5f;
 
 	// Use alternate tick function if we are connected to a server
 	if (m_network != nullptr)
 	{
 		tickNet(playerPos);
 	}
-	else if (m_map != nullptr)
-	{
-		tickLocal(playerPos);
-	}
-	else
-	{
-		assert(m_network != nullptr || m_map != nullptr)
-	}
+	tickLocal(playerPos);
 }
 
 void ChunkView::tickNet(math::vec3 playerPos)
@@ -99,8 +91,8 @@ void ChunkView::tickNet(math::vec3 playerPos)
 	// TODO discard all out of view chunks
 
 	// If we don't already have a chunk, insert it
-	size_t size = m_network->chunkQueue.size();
-	for (size_t i = 0; i < size; i++)
+	std::size_t size = m_network->chunkQueue.size();
+	for (std::size_t i = 0; i < size; i++)
 	{
 		Chunk chunk = Chunk(math::vec3 {});
 		if (!m_network->chunkQueue.try_pop(chunk))
--- a/Phoenix/Client/Source/Graphics/ChunkView.cpp
+++ b/Phoenix/Client/Source/Graphics/ChunkView.cpp
@@ -36,8 +36,8 @@
 using namespace phx::voxels;
 using namespace phx;
 
-ChunkView::ChunkView(int viewDistance, Map&& map)
-    : m_viewDistance(viewDistance), m_map(std::move(map))
+ChunkView::ChunkView(int viewDistance, Map* map)
+    : m_viewDistance(viewDistance), m_map(map)
 {
 	// calculates the maximum visible chunks.
 	const int viewLength       = (viewDistance * 2) + 1;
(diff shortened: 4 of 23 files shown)


Phoenix/Client/Include/Client/Game.hpp
```diff
@@ -112,8 +113,8 @@ namespace phx::client
 		math::vec3  m_prevPos;
 		int         m_playerHand = 0;
 
-		client::Network*    m_network;
-		client::InputQueue* m_inputQueue;
+		client::Network*    m_network    = nullptr;
```
SuperFola, line 116: We could have that in the constructor I think

Phoenix/Client/Include/Client/Game.hpp
```diff
@@ -112,8 +113,8 @@ namespace phx::client
...
+		client::InputQueue* m_inputQueue = nullptr;
```
SuperFola, line 117: In constructor maybe

Phoenix/Client/Include/Client/Network.hpp
```diff
@@ -32,14 +32,15 @@
...
+#include <Common/Voxels/Chunk.hpp>
 #include <thread>
 
 namespace phx::client
 {
 	class Network
 	{
 	public:
-		Network(std::ostringstream& chat);
+		Network(std::ostringstream& chat, phx::net::Address address);
```
SuperFola, line 43: Just wondering, not really a problem, but having the chat as an ostream could be blocking in the future if we want to add coloured Nicknames

Can we pass the address as const ref or is that relatively small and we can afford a copy?

Phoenix/Client/Include/Client/Network.hpp
```diff
@@ -92,9 +95,10 @@ namespace phx::client
 		void sendMessage(std::string message);
 
 		phx::BlockingQueue<std::pair<Position, size_t>> stateQueue;
+		phx::BlockingQueue<voxels::Chunk>               chunkQueue;
 
 	private:
-		bool                m_running;
+		bool                m_running = false;
```
SuperFola, line 101: Maybe have this in constructor

Phoenix/Client/Source/Game.cpp
```diff
@@ -59,10 +59,24 @@ static void rawEcho(const std::string& input, std::ostringstream& cout)
...
+	m_chat = new ui::ChatWindow("Chat Window", 5,
+	                            "Type /help for a command list and help.");
+
+	/// @TODO replace with network callback
+	m_chat->registerCallback(rawEcho);
+
+	if (networked)
+	{
+		m_network = new client::Network(m_chat->cout,
+		                                phx::net::Address("127.0.0.1", 7777));
```
SuperFola, line 74: Just a question, are we planning to have this in the settings.json file? Or whatever to modify it without recompiling

Phoenix/Client/Source/Game.cpp
```diff
@@ -308,18 +325,21 @@ void Game::onAttach()
 	}
 
 	m_inputQueue = new InputQueue(m_registry, m_player, m_camera);
-	m_inputQueue->start(std::chrono::milliseconds(50), m_network);
+	if (m_network != nullptr)
+	{
+		m_inputQueue->start(std::chrono::milliseconds(50), m_network);
```
SuperFola, line 330: Same, can we modify this parameter for slower/faster PC?

Phoenix/Client/Source/Graphics/ChunkView.cpp
```diff
@@ -47,27 +47,105 @@ ChunkView::ChunkView(int viewDistance, Map&& map)
...
+			LOG_WARNING("CHUNK_VIEW")
+			    << "Attempted to pop from empty chunk queue";
+			return;
+		}
+		math::vec3 chunkToCheck = chunk.getChunkPos();
+
+		chunkToCheck = chunkToCheck * static_cast<float>(Chunk::CHUNK_WIDTH);
+
+		auto result = std::find_if(m_activeChunks.begin(), m_activeChunks.end(),
+		                           [chunkToCheck](const Chunk& o) -> bool {
```
SuperFola, line 119: &chunkToCheck, take it by reference to avoid copies

Phoenix/Common/Source/PlayerView.cpp
```diff
@@ -0,0 +1,85 @@
...
+	{
+		for (int y = -viewDistance; y <= viewDistance; y++)
+		{
+			for (int z = -viewDistance; z <= viewDistance; z++)
+			{
+				math::vec3 chunkToCheck = {static_cast<float>(x + posX),
+				                           static_cast<float>(y + posY),
+				                           static_cast<float>(z + posZ)};
+				bool       hasChunk     = false;
+				for (const auto chunk : view.chunks)
```
SuperFola, line 68: Can we have a const auto& chunk or you didn't put a ref because it didn't compile?

Phoenix/Server/Include/Server/Iris.hpp
```diff
@@ -56,6 +57,16 @@ namespace phx::server::net
 		std::string message;
 	};
 
+	struct Event
+	{
+		enum Type
```
SuperFola, line 62: Enum class, will avoid clashes with other names

Phoenix/Server/Include/Server/Iris.hpp
```diff
@@ -153,7 +170,7 @@ namespace phx::server::net
 		BlockingQueue<MessageBundle> messageQueue;
 
 	private:
-		bool                                          m_running;
+		bool                                          m_running = false;
```
SuperFola, line 173: Have this in constructor maybe?

Phoenix/Client/Include/Client/Game.hpp
apachano: Is there any benefit? I was following what other pointers did but I am indifferent here.

Phoenix/Client/Include/Client/Game.hpp
apachano: Same as above

Phoenix/Client/Include/Client/Network.hpp
apachano: You can color over a stream so I don't think that's an issue. This was an easy way to not depend on external systems but I am definitely up for suggestions.

No reason to not pass it as a const ref, it's tiny but doesn't hurt.

Phoenix/Client/Source/Game.cpp
apachano: It'll be a CLI argument on launch

Phoenix/Client/Source/Game.cpp
apachano: Like move it to the config? I'd actually like to derive this from the tick speed, 50ms is 1/20 of a tick, this means we never wait for more than a tick to check for network activity. If we make this larger, then we will see it get caught up and struggle to catch up if packets hit a bump somewhere. Ideally, this is as low as possible to accommodate for the network.

Phoenix/Client/Source/Graphics/ChunkView.cpp
apachano: Agreed, I actually completely rework this in my next PR so can we look at it then? (Otherwise, I just copied existing code)

Phoenix/Common/Source/PlayerView.cpp
apachano: Its an issue with the serializer, I switch this up to pointers in my next PR which reworks the map.

Phoenix/Client/Source/Graphics/ChunkView.cpp
```diff
@@ -47,27 +47,105 @@ ChunkView::ChunkView(int viewDistance, Map&& map)
...
+	{
+		tickNet(playerPos);
+	}
+	else if (m_map != nullptr)
+	{
+		tickLocal(playerPos);
+	}
+	else
+	{
+		LOG_FATAL("ChunkView.cpp")
```
beeperdeeper089, line 93: Rather than this here, can we use an assert to just make sure both aren't nullptr, the likeliness of this should really be impossible, right?

Phoenix/Client/Source/Network.cpp
```diff
@@ -130,6 +137,19 @@ void Network::parseMessage(phx::net::Packet& packet)
 	m_chat << "\n";
 }
 
+void Network::parseData(phx::net::Packet& packet)
+{
+	voxels::Chunk chunk(math::vec3 {0, 0, 0});
```
beeperdeeper089, line 142: just put {} and it will initialize to 0 specifically.

Phoenix/Common/Include/Common/CMakeLists.txt
```diff
@@ -6,11 +6,11 @@ add_subdirectory(Network)
 
 set(currentDir ${CMAKE_CURRENT_LIST_DIR})
 set(Headers
-	${mathHeaders}
-	${voxelHeaders}
-	${cmsHeaders}
-	${serializationHeaders}
-	${networkHeaders}
+        ${mathHeaders}
```
beeperdeeper089, line 9: It has gone from tabs to spaces in this file.

Phoenix/Common/Include/Common/Voxels/Chunk.hpp
```diff
@@ -164,10 +165,11 @@ namespace phx::voxels
 			                      static_cast<std::size_t>(pos.z));
 		}
 
+		Serializer& operator&(Serializer& this_) override;
```
beeperdeeper089, line 168: Rename the parameter to ``serializer`` to match coding standards (or whatever word seems more suitable, ``this`` doesn't work and the underscore after is wrong).

Phoenix/Common/Source/Network/Host.cpp
```diff
@@ -217,7 +217,7 @@ Peer& Host::getPeer(ENetPeer& peer)
 	{
 		return m_peers.at(std::size_t(peer.data));
 	}
-	std::cout << "peer not found";
+	LOG_WARNING("NETCODE") << "peer not found";
```
beeperdeeper089, line 220: Capitalize to ``Peer not found.`` and/or make this a bit more descriptive so an enduser can half understand this.

Phoenix/Common/Source/PlayerView.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+	// this gets the raw player position in voxel-world coordinates.
+	math::vec3 playerPos =
+	    (registry->get<Position>(entity).position / 2.f) + 0.5f;
+	const int posX = static_cast<int>(playerPos.x) / voxels::Chunk::CHUNK_WIDTH;
+	const int posY =
+	    static_cast<int>(playerPos.y) / voxels::Chunk::CHUNK_HEIGHT;
+	const int posZ = static_cast<int>(playerPos.z) / voxels::Chunk::CHUNK_DEPTH;
+
+	const int viewDistance = 3;
```
beeperdeeper089, line 54: Are we allowing this to be variable?

Phoenix/Common/Source/Voxels/Chunk.cpp
```diff
@@ -99,3 +99,33 @@ void Chunk::setBlockAt(phx::math::vec3 position, BlockType* newBlock)
 		m_blocks[getVectorIndex(position)] = newBlock;
 	}
 }
+
+phx::Serializer& Chunk::operator&(phx::Serializer& this_)
```
beeperdeeper089, line 103: Parameter name as said above.

Phoenix/Server/Source/Game.cpp
```diff
@@ -50,32 +54,73 @@ void Game::run()
...
+				auto entity = m_registry->get<Player>(event.player);
+				for (const auto& chunk :
+				     PlayerView::update(m_registry, entity.actor, &m_map))
+				{
+					m_iris->sendData(entity.id, chunk);
+				}
+				break;
+			}
+			default:
+				LOG_WARNING("GAME") << "Event received with unknown type";
```
beeperdeeper089, line 110: Replace with a more descriptive error message.

beeperdeeper089: Also put a break statement at the end of this (you don't need it but do it anyway for the sake of consistency)

Phoenix/Common/Include/Common/Voxels/Chunk.hpp
apachano: Since it's overriding the underlying function, don't the parameter names need to match?

Phoenix/Common/Source/PlayerView.cpp
apachano: Eventually yes, needs a todo to move this to a config (The server should limit what the max view distance is and the client should have a separate render distance). This needs built out further than what I have.

Phoenix/Server/Source/Game.cpp
apachano: How would you describe it? There really isn't any additional context here that I can think of

Phoenix/Common/Source/Network/Host.cpp
beeperdeeper089: Also, we are reaching the end of a function which needs a return.

Phoenix/Common/Include/Common/Voxels/Chunk.hpp
SuperFola: No, it doesn't need to, only the types (+ const, ref, pointers) need to match

Phoenix/Client/Source/Graphics/ChunkView.cpp
SuperFola: Okay, np

Phoenix/Server/Source/Game.cpp
beeperdeeper089: I would just say ``Invalid network event received``, idk it sounds better than ``unknown``

Phoenix/Common/Source/PlayerView.cpp
beeperdeeper089: Ayt, can you put a ``/// @todo`` above it so we can find it when we need to.

Phoenix/Client/Include/Client/Game.hpp
apachano: Holding on changes pending https://github.com/GentenStudios/Genten/issues/4

Phoenix/Client/Include/Client/Network.hpp
apachano: Holding pending https://github.com/GentenStudios/Genten/issues/4

Phoenix/Client/Source/Graphics/ChunkView.cpp
apachano: I upgraded to an assert, I also completly wipe this in my next PR and build it into the map instead. Realy its not possible to hit this, but was a catch since I did a conditional on the else.

Phoenix/Common/Include/Common/CMakeLists.txt
apachano: Hmm 🤔 

I fixed this in my settings to auto tab instead of space, but the default is spaces. I'm also interested if we can just configure the clang-format to cover this (the IDE warns that it can)

Phoenix/Common/Source/Network/Host.cpp
apachano: Fixed the capitalization, I'm not sure what to do for better descriptions. The enduser should never see this TBH but we need to prevent the exception from randomly throwing. If you are trying to diagnose at this level, you can cmdF the file to find the message.

Phoenix/Common/Source/PlayerView.cpp
apachano: No, for the usage in this PR it should be in a config, I do some stuff in the next PR that means it needs to be a parameter. Really, I have no clue how we are going to supply this since on the server its a global config and on the client its a setting (both use the same class)

I slapped a todo on it and I will think it over.

Phoenix/Server/Source/Game.cpp
apachano: Added the break,

IDK about the error message, "Unexpected event type received" is a bit less verbose. There really isn't anything going on here and the event is an ENUM so you need a debugger to diagnose what's happening. Also this should never happen, clang-format just wanted a default clause.

apachano:
@beeperdeeper089 Re-review plz

Phoenix/Server/Source/Game.cpp
apachano: Love that GH only auto-refreshes some times

Phoenix/Common/Source/Network/Host.cpp
apachano: Hmm, yea we do reach the end of a function that needs a return 🤔  what do we return if there is nothing to return? Maybe we swap the return to a pointer?

beeperdeeper089: Yeah, I would assume a pointer is the way to go.

beeperdeeper089:
Replace the peer bit with a pointer and return nullptr if not found, once we get we good to merge

Phoenix/Client/Source/Graphics/ChunkView.cpp
```diff
@@ -47,27 +47,103 @@ ChunkView::ChunkView(int viewDistance, Map&& map)
...
+	m_renderer->buildTextureArray();
+}
+
+ChunkView::~ChunkView()
+{
+	delete m_renderer;
+	delete m_map;
+}
 
 void ChunkView::tick(math::vec3 playerPos)
```
SuperFola, line 67: i must have missed this one, we can have a const ref here for `playerPos`

Phoenix/Client/Source/Graphics/ChunkView.cpp
```diff
@@ -47,27 +47,103 @@ ChunkView::ChunkView(int viewDistance, Map&& map)
...
+	{
+		if (m_network->chunkQueue.empty())
+		{
+			return;
+		}
+	}
+
+	// this converts the raw camera/player position into voxel-world positions.
 	playerPos = playerPos / 2.f;
 	playerPos += 0.5f;
```
SuperFola, line 80: maybe less readable for some people but i think it's better to write it like `playerPos = (playerPos / 2.f) + 0.5f;`

Phoenix/Client/Source/Graphics/ChunkView.cpp
```diff
@@ -47,27 +47,103 @@ ChunkView::ChunkView(int viewDistance, Map&& map)
...
+	{
+		tickNet(playerPos);
+	}
+	else if (m_map != nullptr)
+	{
+		tickLocal(playerPos);
+	}
+	else
+	{
+		assert(m_network != nullptr || m_map != nullptr)
```
SuperFola, line 93: the assert will only work in debug mode, is it wanted?

Phoenix/Client/Source/Graphics/ChunkView.cpp
```diff
@@ -47,27 +47,103 @@ ChunkView::ChunkView(int viewDistance, Map&& map)
...
+	}
+}
+
+void ChunkView::tickNet(math::vec3 playerPos)
+{
+	// TODO discard all out of view chunks
+
+	// If we don't already have a chunk, insert it
+	size_t size = m_network->chunkQueue.size();
```
SuperFola, line 102: std::size_t, even though size_t works, but for consistency using the std:: version would be better (iirc we had that corrected in another PR, and it would be better to use the same version everywhere, correct me if i'm wrong)

Phoenix/Common/Source/Voxels/Chunk.cpp
```diff
@@ -99,3 +99,33 @@ void Chunk::setBlockAt(phx::math::vec3 position, BlockType* newBlock)
...
 	}
 }
+
+phx::Serializer& Chunk::operator&(phx::Serializer& ser)
+{
+	if (ser.m_mode == Serializer::Mode::WRITE)
+	{
+		ser& m_pos.x;
+		ser& m_pos.y;
+		ser& m_pos.z;
```
SuperFola, line 109: won't `ser & m_pos.x & m_pos.y & m_pos.z` work?

Phoenix/Common/Source/Voxels/Chunk.cpp
```diff
@@ -99,3 +99,33 @@ void Chunk::setBlockAt(phx::math::vec3 position, BlockType* newBlock)
...
+	else
+	{
+		m_blocks.clear();
+		ser& m_pos.x;
+		ser& m_pos.y;
+		ser& m_pos.z;
+		for (int i = 0; i < CHUNK_DEPTH * CHUNK_WIDTH * CHUNK_HEIGHT; i++)
+		{
+			size_t id;
+			ser&   id;
```
SuperFola, line 126: is it an error? id will most likely be uninitialized here, or is the serializer pushing a value into it?

Phoenix/Server/Source/Game.cpp
```diff
@@ -50,32 +54,74 @@ void Game::run()
...
+			    static_cast<int>(pos.x) / voxels::Chunk::CHUNK_WIDTH,
+			    static_cast<int>(pos.y) / voxels::Chunk::CHUNK_HEIGHT,
+			    static_cast<int>(pos.z) / voxels::Chunk::CHUNK_DEPTH);
+			ActorSystem::tick(m_registry, player.actor, dt, state.second);
+			pos = m_registry->get<Position>(player.actor).position;
+			const math::vec3i newPos(
+			    static_cast<int>(pos.x) / voxels::Chunk::CHUNK_WIDTH,
+			    static_cast<int>(pos.y) / voxels::Chunk::CHUNK_HEIGHT,
+			    static_cast<int>(pos.z) / voxels::Chunk::CHUNK_DEPTH);
+			if (!(oldPos == newPos))
```
SuperFola, line 81: if oldPos != newPos

apachano: Doesn't work. The logic for != isn't !( == ) so it comes out with data I didn't want.

Phoenix/Common/Source/Voxels/Chunk.cpp
apachano: This is the serializer

Phoenix/Client/Source/Graphics/ChunkView.cpp
apachano: @beeperdeeper089 ??? Thoughts, you wanted this to be an assert

apachano: @beeperdeeper089 Any thoughts? You wanted the assert

beeperdeeper089: Is there a possibility of this happening during runtime? (I would assume not)

If not, I meant something along the lines of:
```cpp
// make sure both aren't nullptr, we only need this in debug mode because this probably won't happen in a release build
// if it doesn't happen in a debug build cos we won't have had this happen.
assert(m_network != nullptr && m_map != nullptr)

if (m_network)
{
	tickNet(playerPos);
}
else
{
    tickLocal(playerPos);
}
```

apachano: Gotcha, I can change it to that then (or do we even care about asserting in a situation like this? We will fault when we try to call a function on a null pointer anyways)

beeperdeeper089: Right, if you know for sure that this cannot happen then drop the assert, otherwise keep it so we can get better traceability in debug builds.

Phoenix/Server/Source/Game.cpp
beeperdeeper089: I thought the logic for ``!=`` is the same as ``!(==)``. The ``!=`` just inverts a bit on equality afaik.

SuperFola: If it doesn't than we have a big problem houston, our vector class sucks and is not coded as we need it to be

apachano: The vector class likely needs some help. I haven't dug into it yet, the last time I did I left with a headache. We can toss a ticket in the backlog to investigate but I would argue that is out of scope for this PR.

beeperdeeper089: Yup, sounds good.

Phoenix/Client/Source/Graphics/ChunkView.cpp
apachano: We modify the value in the function. Unless we create a new variable in the body this needs to create a copy here.

PR #265, opened 2020-07-14
Revised save handling system

beeperdeeper089:
Resolves: #
Authors:
@beeperdeeper089 

## Summary of changes
- Use JSON to store mod list and name of save
- Single collective class to load mod list and eventually other settings, this will come in more useful once we implement simple dimensions/mapgen.
  
## Caveats
- Not currently implemented due to a necessity to refactor the Map class.

## On approval
Wait for me, I might be able to pull the two PRs in at a similar time.

### DIFF
--- a/Phoenix/Client/Source/Game.cpp
+++ b/Phoenix/Client/Source/Game.cpp
@@ -81,20 +81,27 @@ Game::Game(gfx::Window* window, entt::registry* registry, bool networked)
 	std::fstream             fileStream;
 	std::vector<std::string> toLoad;
 
-	fileStream.open("Saves/" + save + "/Mods.txt");
-	if (!fileStream.is_open())
-	{
-		LOG_FATAL("CMS") << "Error opening save file";
-		exit(EXIT_FAILURE);
-	}
+	auto saveToUse = "save1234";
+	auto existingSaves = Save::listAllSaves();
 
-	std::string input;
-	while (std::getline(fileStream, input))
+	if (std::find(existingSaves.begin(), existingSaves.end(), saveToUse) == existingSaves.end())
 	{
-		toLoad.push_back(input);
-	}
+		// gotta create a save, so we gotta give it a mod list too.
 
-	m_modManager = new cms::ModManager(toLoad, {"Modules"});
+		// we should use a mod list command line argument, but in the meantime
+		// lets hard code it.
+
+		std::vector<std::string> mods = {"mod1", "mod2", "mod3"};
+
+		m_save = new Save(saveToUse, mods);
+	}
+	else
+	{
+		// it exists so lets load it.
+		m_save = new Save(saveToUse);
+	}
+	
+	m_modManager = new cms::ModManager(m_save->getModList(), {"Modules"});
 
 	voxels::BlockRegistry::get()->registerAPI(m_modManager);
 
--- a/Phoenix/Client/Source/Main.cpp
+++ b/Phoenix/Client/Source/Main.cpp
@@ -28,12 +28,19 @@
 
 #include <Client/Client.hpp>
 
+#include <Common/Save.hpp>
+
 using namespace phx;
 
 #undef main
 int main(int argc, char** argv)
 {
-	client::Client::get()->run();
+	//client::Client::get()->run();
+
+	auto save = phx::Save::createSave({"bob", {}, {{"shrek", "pizza"}}});
+
+	save.toFile();
+	save.toFile("bob2");
 	
 	return 0;
 }
--- a/Phoenix/Client/Source/Main.cpp
+++ b/Phoenix/Client/Source/Main.cpp
@@ -28,20 +28,12 @@
 
 #include <Client/Client.hpp>
 
-#include <Common/Save.hpp>
-
 using namespace phx;
 
 #undef main
 int main(int argc, char** argv)
 {
-	//client::Client::get()->run();
-
-	auto saves = Save::listAllSaves();
-	for (auto& save : saves)
-	{
-		std::cout << save << std::endl;
-	}
+	client::Client::get()->run();
 
 	return 0;
 }
--- a/Phoenix/Client/Source/Main.cpp
+++ b/Phoenix/Client/Source/Main.cpp
@@ -28,11 +28,20 @@
 
 #include <Client/Client.hpp>
 
+#include <Common/Save.hpp>
+
 using namespace phx;
 
 #undef main
 int main(int argc, char** argv)
 {
-	client::Client::get()->run();	
+	//client::Client::get()->run();
+
+	auto saves = Save::listAllSaves();
+	for (auto& save : saves)
+	{
+		std::cout << save << std::endl;
+	}
+
 	return 0;
 }
--- a/Phoenix/Client/Source/Main.cpp
+++ b/Phoenix/Client/Source/Main.cpp
@@ -28,23 +28,12 @@
 
 #include <Client/Client.hpp>
 
-#include <Common/Logger.hpp>
-
-#include <Common/Save.hpp>
-
 using namespace phx;
 
 #undef main
 int main(int argc, char** argv)
 {
-	//client::Client::get()->run();
-
-	Logger::initialize({});
-
-	phx::SaveConfig config;
-	config.name = "epic gamer world";
-	config.mods = {"mod1", "mod2"};
-	phx::Save::createSave(config);
+	client::Client::get()->run();
 	
 	return 0;
 }
--- a/Phoenix/Client/Source/Main.cpp
+++ b/Phoenix/Client/Source/Main.cpp
@@ -28,19 +28,11 @@
 
 #include <Client/Client.hpp>
 
-#include <Common/Save.hpp>
-
 using namespace phx;
 
 #undef main
 int main(int argc, char** argv)
 {
-	//client::Client::get()->run();
-
-	auto save = phx::Save::createSave({"bob", {}, {{"shrek", "pizza"}}});
-
-	save.toFile();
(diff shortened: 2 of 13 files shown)


beeperdeeper089:
well, that's filesystem being a bitch...

Phoenix/Common/Include/Common/Save.hpp
```diff
@@ -0,0 +1,84 @@
...
+		nlohmann::json settings;
+	};
+
+	class Save
+	{
+	public:
+		Save(const std::string& save);
+		~Save();
+
+		static Save createSave(const SaveConfig& config);
```
SuperFola, line 63: do we want to add documentation here to be explicit or does it speak for itself?

Phoenix/Common/Source/Save.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+using namespace phx;
+
+Save::Save(const std::string& save)
+{
+	namespace fs = std::filesystem;
+
+	auto path = fs::current_path();
+	path /= "Saves";
+	path /= save;
```
SuperFola, line 44: can be simplified as `auto path = fs::current_path() / "Saves" / save;`

Phoenix/Common/Source/Save.cpp
```diff
@@ -0,0 +1,113 @@
...
+{
+}
+
+Save Save::createSave(const SaveConfig& config)
+{
+	namespace fs = std::filesystem;
+
+	auto path = fs::current_path();
+	path /= "Saves";
+	path /= config.name;
```
SuperFola, line 90: can be simplified

Phoenix/Common/Include/Common/Save.hpp
beeperdeeper089: I'll document it just so we don't have to end up doing it later.

Phoenix/Common/Source/Save.cpp
beeperdeeper089: Yeah, I saw you could do that just now lol

Phoenix/Common/Source/Input.cpp
```diff
@@ -28,7 +28,7 @@
 
 #include <Common/Input.hpp>
 
-phx::Serializer& phx::InputState::operator&(phx::Serializer& this_)
+phx::Serializer& phx::InputState::operator&(phx::Serializer& serializer)
```
apachano, line 31: We name this "ser" everywhere else, its a bit lazy but it does the job. Do we care about consistency here? If so, which direction should we go?

Phoenix/Common/Source/Save.cpp
```diff
@@ -0,0 +1,109 @@
...
+		exit(EXIT_FAILURE);
+	}
+
+	json >> m_data.settings;
+	if (!m_data.settings["mods"].is_array())
+	{
+		LOG_FATAL("SAVES") << "The save: " << save
+		                   << " does not have a valid configuration, quitting.";
+		exit(EXIT_FAILURE);
+	}
```
apachano, line 65: Am I reading this right in that this error will be thrown if no mods are listed in the config file? If so, can we state that explicitly in the logged message instead of copying the above error?

Phoenix/Common/Source/Save.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+	m_data.mods = m_data.settings["mods"].get<std::vector<std::string>>();
+	m_data.name = m_data.settings["name"].get<std::string>();
+
+	LOG_INFO("SAVES") << "The save: " << m_data.name << " requires the mods: ";
+	for (const std::string& mod : m_data.mods)
+	{
+		LOG_INFO("SAVES") << mod;
+	}
```
apachano, line 74: IMO, and push back here, this is a bit excessive to log. Once you hit say 100 mods this will just fill the terminal/ logs and provides no information on if the mods loaded. 

I would argue that we remove it or possible only log to debug?

Phoenix/Common/Include/Common/Save.hpp
```diff
@@ -0,0 +1,121 @@
...
+		//Dimension* createDimension(const std::string& name);
+		//Dimension* getDimension(const std::string& name);
+		//void       setDefaultDimension(Dimension* dimension);
+		//Dimension* getDefaultDimension();
+		//const std::vector<std::string>& getDimensions() const;
+
+		/**
+		 * @brief Saves everything to be saved in the Save.
+		 */
+		void save();
```
apachano, line 113: I don't have an answer for you here so maybe you just ignore me, but having the function named the same as the class is hard to look at IMO, is there something else we can name this?

Phoenix/Common/Source/Input.cpp
beeperdeeper089: idm, ``ser`` should work fine, lets stick to that.

Phoenix/Common/Source/Save.cpp
beeperdeeper089: Even if creating a save without any mods, the JSON will save with an array, thus if there isn't an array, it's broken.

Phoenix/Common/Source/Save.cpp
beeperdeeper089: Oh yeah, this was a testing measure, forgot to remove it.

Phoenix/Common/Include/Common/Save.hpp
beeperdeeper089: I agree here, I spent a few minutes just tryna think of what to call it but ``save`` just felt like the only real option in this scenario.

SuperFola: maybe `toFile(optionnal name)` or `create`?

apachano: Serious +1 for toFile()

It makes sense and is 1000x better than mimicking the class name.

beeperdeeper089: idk about ``toFile``, put *what* to file, what about ``write``, that's probably just as ambiguous but "writing" a save sounds better to my ears

SuperFola: `save.toFile` => save to file (the save/data)

but yeah, `save.write(optional name)` would be great as well

Phoenix/Common/Include/Common/Save.hpp
```diff
@@ -0,0 +1,137 @@
...
+		~Save();
+
+		/**
+		 * @brief Creates a new save if one does not exist.
+		 * @param config The configuration for the new save.
+		 * @return The new save.
+		 *
+		 * If the save already exists it will return that save.
+		 */
+		static Save createSave(const SaveConfig& config);
```
apachano, line 93: Just commenting in relation to above, why don't we just make this a constructor?

Phoenix/Client/Source/Main.cpp
```diff
@@ -28,17 +28,19 @@
...
 
 using namespace phx;
 
 #undef main
 int main(int argc, char** argv)
 {
-	client::Client::get()->run();
+	//client::Client::get()->run();
+
+	auto save = phx::Save::createSave({"bob", {}, {{"shrek", "pizza"}}});
```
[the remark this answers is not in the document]
beeperdeeper089: I think this is fine for ``createSave``, since a lot of functions in other contexts would make sense to return the existing object instead of creating on in a similar scenario.

Phoenix/Common/Include/Common/Save.hpp
beeperdeeper089: I would say that creating a save should be explicit, and if you choose a save that doesn't exist then we'll just tell you the save doesn't exist, make one. (imo, feel free to argue this)

apachano: Agreed, but this could still be a constructor instead of a static function. I'm okay leaving it as is, but its also funky IMO since creating a save object (regardless of how it works internally) sounds like a constructor.

Phoenix/Client/Source/Game.cpp
```diff
@@ -81,20 +81,27 @@ Game::Game(gfx::Window* window, entt::registry* registry, bool networked)
 	std::fstream             fileStream;
 	std::vector<std::string> toLoad;
 
-	fileStream.open("Saves/" + save + "/Mods.txt");
-	if (!fileStream.is_open())
+	auto saveToUse = "save1234";
+	auto existingSaves = Save::listAllSaves();
+
+	if (std::find(existingSaves.begin(), existingSaves.end(), saveToUse) == existingSaves.end())
```
apachano, line 87: This is redundant. The Save constructor itself checks if the save already exists so if we just call m_save = new Save(saveToUse, {"mod1", "mod2", "mod3"}) then it will work fine. I would argue that instead of doing the extra check, pass the mod list every time regardless and the constructor can discard it if the save already exists. An empty vector when we know we don't need a new save should be fine.

Phoenix/Common/Include/Common/Save.hpp
```diff
@@ -0,0 +1,144 @@
...
+	{
+		// name of save.
+		std::string name;
+
+		// mod list, passed as CLI arguments.
+		std::vector<std::string> mods;
+
+		// save specific settings, etc...
+		nlohmann::json settings;
+	};
```
apachano, line 63: We read these values one at a time from the json and we don't actually pass the object anywhere. Why not just store these three objects in the main class independently?

beeperdeeper089: Yeah, it was supposed to be passed around everywhere and ended up not needed, I'll merge for now but I'll just get rid of it tomorrow

Issue #269, opened 2020-07-25
Segfault After Clicking on the Blank Game Screen


zicklag:
## Context

I'm building the game in a "Docker" container on my host using [udocker](https://github.com/indigo-dc/udocker) because I don't have root access on my build machine to install dependencies using `apt` so this may not be a bug worth addressing unless you can reproduce it on your own machine.

I used an Ubuntu 20.04 base docker image and installed CMake, g++, and the packages that Cmake complained about needing during the build.

**Commit:** ece10094e6436b

## Describe the bug

Segfault when clicking after the game appears to have loaded, but there is nothing on the screen other than imgui and the crosshair.

## To Reproduce

When starting up the game I:

- Am presented with the Genten Studio logo splash image
- The game starts and I can see the imgui promp and a screen crosshair with now graphics behind it:

![image](https://user-images.githubusercontent.com/25393315/88465680-554ade00-ce8a-11ea-8559-243ea1e1df32.png)

- As soon as I click the program segfaults:

**Log:**

```
[INFO][LOGGING] A log file was not specified, logging only to console.
[INFO][MODULE] Load mod 1
[INFO][MODULE] Load mod 3
[INFO][MODULE] Load mod 2
[INFO][MAIN] Registering world
[INFO][MAIN] Prepare rendering
[INFO][MAIN] Register GUI
[INFO][MAIN] Game layer attached
```

**Stacktrace:**

```
phx::ActorSystem::action1(entt::basic_registry<entt::entity>*, entt::entity) (/home/zicklag/git/other/Phoenix/Phoenix/ThirdParty/entt/src/entt/entity/sparse_set.hpp:357)
phx::client::Game::onEvent(phx::events::Event&) (/home/zicklag/git/other/Phoenix/Phoenix/Client/Source/Game.cpp:410)
phx::gfx::LayerStack::onEvent(phx::events::Event) (/home/zicklag/git/other/Phoenix/Phoenix/Client/Source/Graphics/LayerStack.cpp:159)
phx::client::Client::onEvent(phx::events::Event) (/home/zicklag/git/other/Phoenix/Phoenix/Client/Source/Client.cpp:115)
phx::gfx::Window::dispatchToListeners(phx::events::Event&) (/home/zicklag/git/other/Phoenix/Phoenix/Client/Source/Graphics/Window.cpp:323)
[...]

PR #275, opened 2020-08-05
Reimplement Chat functionality with a few more features.

beeperdeeper089:
Resolves: #269 (i think will help mitigate it)
Authors:
@beeperdeeper089 

## Summary of changes
- Reimplement the chat class with a box at the top.
- Add "invisible" messages (show messages even with chat not being open).
- Add easier way to send messages, automatically pushes new message on enter.
- Removes the funky function pointer mess with the old Chat functionality.
- Fix bug with meshing
- Fix crash with chunks not actually being generated.
  
## Caveats
- None that I know of

## On approval
Merge.


tobyplowy:
@beeperdeeper089 Is this still a draft? You did edit it like an hour ago and you marked it as merge after approval so it's a little confusing.

beeperdeeper089:
how tf you change a draft pr to a normal pr

beeperdeeper089:
oh nvm theres a fucking button

Phoenix/Client/Include/Client/Network.hpp
```diff
@@ -43,7 +45,7 @@ namespace phx::client
 	class Network
 	{
 	public:
-		Network(std::ostringstream& chat, const phx::net::Address& address);
+		Network(const phx::net::Address& address, gfx::ChatBox* chat);
```
apachano, line 48: I'mma push back here that we should either go in the opposite direction or stick to a stream. This makes the networking rely explicitly on the chat box vs a generic receiver for data. I would argue that we pass the networking to the chatbox which subscribes instead to match the pattern of other systems.

Phoenix/Client/Include/Client/Graphics/ChatBox.hpp
```diff
@@ -28,36 +28,53 @@
...
-		inline void focus() { m_renderFocus = true; };
+		void tick(float dt);
 
-		void draw(bool* p_open = nullptr, ImGuiWindowFlags flags = ImGuiWindowFlags_None);
-	};
-}; // namespace UI
+		void setMessageCallback(
+		    const std::function<void(const std::string& message)>& callback);
+
+		void pushMessage(const std::string& message);
```
apachano, line 56: Hmm :thinking: Maybe printMessage instead? IDK why its throwing me off when there is less context and I feel like we are pushing the message to the server. 

Maybe I'm crazy tho, not a blocker

Phoenix/Client/Source/Graphics/ChatBox.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+		ImGui::PopStyleColor();
+	}
+
+	// clang-format on
+}
+
+void ChatBox::tick(float dt)
+{
+	// limit history to 30, can be increased in the future.
```
apachano, line 142: Can we just store this in a setting? If not lets move the magic number to a const

Phoenix/Client/Include/Client/Network.hpp
beeperdeeper089: Yeah, that was my preferred design, but this was the least "intrusive" way to change it. I'll switch it to the chat taking a network.

Phoenix/Client/Include/Client/Graphics/ChatBox.hpp
beeperdeeper089: Hmm, i think pushMessage makes more sense, since it's pushing a message onto a queue, I'm fine with changing it, but because it's not printing to console or printing to just one final place (both just the screen and the invisible box), I'm hesitant to call it ``print``ing

Phoenix/Client/Source/Graphics/ChatBox.cpp
beeperdeeper089: Yeah, forgot about settings, I'll change it over.

Phoenix/Client/Include/Client/Graphics/ChatBox.hpp
apachano: Makes sense, I'm not sure here so leave it.

The design as a whole is kinda funky IMO

Phoenix/Client/Source/Graphics/ChatBox.cpp
beeperdeeper089: Actually, since the Setting is a pointer to a completely different place in memory, it may not be a great idea to have that in a render loop since that could be bad in terms of caching, I think i might stick to a const for now until we might have a better idea for memory layout in the settings system.

PR #281, opened 2020-08-19
Implement Map Events

beeperdeeper089:
Resolves: #
Authors:
@beeperdeeper089 

## Summary of changes
- Implements a map subscriber system.
- Fixes the issue with chunk updates when block break/places occur.
  
## Caveats
- May result in some overhead, however blocking queue's may help fix that issue - we may need to improve cache efficiency with this later (my frame time graph showed essentially no changes so I think we're fine right now).

## On approval
Merge.


Phoenix/Client/Include/Client/Graphics/ChunkMesher.hpp
```diff
@@ -40,6 +40,7 @@
 
 #include <Common/Math/Math.hpp>
 #include <Common/Voxels/Block.hpp>
+#include <Common/Voxels/Chunk.hpp>
```
apachano, line 43: Chunk.hpp includes Block.hpp - we can remove block here

Phoenix/Client/Include/Client/Voxels/BlockRegistry.hpp
```diff
@@ -76,19 +73,19 @@ namespace phx::client
...
 				    {
 					    block.category = voxels::BlockCategory::LIQUID;
 				    }
 				    else
 				    {
-					    // default to air if not liquid or solid.
-					    block.category = voxels::BlockCategory::AIR;
+					    // default to solid if not liquid or air.
+					    block.category = voxels::BlockCategory::SOLID;
 				    }
```
apachano, line 89: Why flip these? Since most blocks are Solid wouldn't it be more performant to check that first? If not - you missed updating a comment

apachano (changes requested):
Minor stuff, otherwise this all looks good

Phoenix/Client/Include/Client/Voxels/BlockRegistry.hpp
beeperdeeper089: Ah yes, I forgot to update the comment, but essentially the intention was to default to Solid in the case that the input is invalid or one is not selected but what might be a smarter idea is:
```cpp
block.category = voxels::BlockCategory::SOLID
if (category == "Solid") ...
if (category == "Liquid") ...
if (category == "Air") ...
```

Phoenix/Client/Include/Client/Graphics/ChunkMesher.hpp
beeperdeeper089: The only reason I kept ``Block.hpp`` is because we are referencing ``voxels::BlockType`` directly, we don't need to but I feel that since we're using that exclusively too, it's worth including.

My opinion is that if we are exclusively using a specific class we should include it regardless - this may be a discussion worth having for coding standards as a whole.

beeperdeeper089: Though, in this case we can move this include directly to the ``.cpp`` file since the header doesn't need it.

apachano: Definitely move it to the .cpp header - how do modern compilers handle redundant includes? My understanding of C is if you include it twice it actually builds the file twice which we don't want to happen.

Phoenix/Client/Include/Client/Voxels/BlockRegistry.hpp
apachano: That makes sense - although now that I think about it by defaulting to Solid that means you don't have to define it in the Lua which reduces size and complexity there. In which case - if you dont define in Lua then it does all three checks in the CMS and we might as well go with what you did + remove the comment for the most optimized operation.

Phoenix/Client/Include/Client/Graphics/ChunkMesher.hpp
beeperdeeper089: That's where the include guards/``#pragma once`` come in, they just make sure code isn't included twice since it might not only build twice but it'll end up failing since you've got multiple redefinitions, etc...

beeperdeeper089:
sol2 strikes once again 😎

apachano (approved):

PR #282, opened 2020-08-19
Merge changes to top level CMakeLists.txt

JosiahWI:
Resolves: N/A
Authors:
@josiah_wi

## Summary of changes
Automatically update git submodules during build.
Change minimum CMake version to 3.12.
  
## On approval
Merge

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -44,9 +44,7 @@ jobs:
 #        key: 'ccache | "$(Agent.JobName)"'
 #        path: $(CCACHE_DIR)
 #      displayName: 'CCache'
-    - script: git submodule update --init
-      displayName: 'Cloning Submodules'
-    - script: cmake -H. -BBuild
+    - script: cmake -S . -B Build -D UPDATE_SUBMODULES=ON
       displayName: 'CMake Configure'
     - script: cmake --build Build
       displayName: 'CMake Build'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -44,7 +44,7 @@ jobs:
 #        key: 'ccache | "$(Agent.JobName)"'
 #        path: $(CCACHE_DIR)
 #      displayName: 'CCache'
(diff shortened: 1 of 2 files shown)


beeperdeeper089:
I'll take a look at this, and test it with CMake 3.18.1 (I use 3.16 rn but updating is always handy).

If we're cloning in the CMake, we should probably remove the part from the CI script.

beeperdeeper089:
@JosiahWI CMake 3.18.1 works just fine, so I would remove the version range within the CMake. Just set the minimum and that should be fine. The Git thing should be fine, but i would say change the messages to:
```
Updating Git Submodules
and
Git Submodules will not be updated, manual checking out is required.
```
or something more descriptive than ``Not updating``

JosiahWI:
The version range sets the correct policies if the CMake version is 3.12 or newer.

beeperdeeper089:
```
cmake_minimum_required(VERSION 3.8...3.16)
```
Right, but this line can become:
```
cmake_minimum_required(VERSION 3.8)
```

azure-pipelines.yml
```diff
@@ -44,9 +44,7 @@ jobs:
 #        key: 'ccache | "$(Agent.JobName)"'
 #        path: $(CCACHE_DIR)
 #      displayName: 'CCache'
-    - script: git submodule update --init
-      displayName: 'Cloning Submodules'
-    - script: cmake -H. -BBuild
+    - script: cmake -S . -B Build -D UPDATE_SUBMODULES=ON
```
beeperdeeper089, line 47: If the option is on by default, do we need to define this?

Also, it's recommended to not use spaces in the CMake command, so like:
```
cmake -S. -BBuild -DUPDATE_SUBMODULES=ON
```

beeperdeeper089: Ignore, me, the spaces is something that was an issue a while ago with things like Command Prompt on Windows which wouldn't work with spaces for some reason.

beeperdeeper089 (approved):

PR #284, opened 2020-08-20
More changes to CMake build.

JosiahWI:
Resolves: Error linking <filesystem> on GCC 8.x
Authors:
@JosiahWI 

## Summary of changes
- CMake links std++fs library if GCC version < 9.x
- CMake only sets compile definitions on relevant third party libraries.
  
## Caveats

## On approval
Wait for confirmation.


Phoenix/ThirdParty/CMakeLists.txt
```diff
@@ -18,17 +15,22 @@ set(ALSOFT_TESTS OFF)
...
+
 add_subdirectory (Glad)
 add_subdirectory (ImGui)
 add_subdirectory (sol2)
 add_subdirectory (lua)
 add_subdirectory (enet)
 add_subdirectory (entt)
+
+# NOMINMAX setting required for std::min std::max to work properly on Windows.
+target_compile_definitions(EnTT INTERFACE $<$<PLATFORM_ID:Windows>:NOMINMAX>)
```
beeperdeeper089, line 32: We will probably actually need ``NOMINMAX`` across the whole project, since we want to use std::min and std::max in a few places.

beeperdeeper089 (changes requested):

Phoenix/Common/CMakeLists.txt
```diff
@@ -4,10 +4,14 @@ add_subdirectory(Include/Common)
 add_subdirectory(Source)
 
 add_library(${PROJECT_NAME} STATIC ${Headers} ${Sources})
-target_link_libraries(${PROJECT_NAME} PRIVATE ${PHX_THIRD_PARTY_LIBRARIES})
+target_link_libraries(${PROJECT_NAME} 
+                      PRIVATE ${PHX_THIRD_PARTY_LIBRARIES}
+                      # Link <filesystem> for versions of GCC < 9.0
+                      PRIVATE $<$<AND:$<CXX_COMPILER_ID:GNU>,$<VERSION_LESS:$<CXX_COMPILER_VERSION>,9.0>>:stdc++fs>)
```
beeperdeeper089, line 10: We're gonna need filesystem on all targets eventually, one of my PRs is using it in the client right now, and the server will end up needing it sooner or later, so I would recommend we put this in the ThirdParty CMake.

Another problem that this does not address is MacOS, where anything before 10.15 does NOT support filesystem regardless of what we tried (iirc), so we would need to test for that and make sure that works, or limit OS compatibility dependent on the version (if CMake can read that) - or figure out something for the package of clang that Mac's use. (I'm not a mac expert, nor do i own a mac so take this with a pinch of salt, @apachano would be the person to consult.)

beeperdeeper089 (changes requested):

Phoenix/ThirdParty/CMakeLists.txt
JosiahWI: I think this is a better case for using add_definitions(), but not a great one. I'm trying to focus my suggested changes around the modern CMake design, which focuses heavily on using targets so that things are encapsulated. The equivalent C/++ code for add_definitions() would be a `#define NOMINMAX` in every single source file, including those in third party libraries.

Phoenix/Common/CMakeLists.txt
JosiahWI: CMake can test whether it works (try_compile()). I think it can test the OS version too, but it's been a bit since I looked at the documentation for those variables.

JosiahWI: If we did use a global command for this, why would it go in ThirdParty? ThirdParty doesn't use <filesystem> and someone trying to maintain the build system won't have a clue where to find it or why it's there.

beeperdeeper089: That's why you would comment code, my opinion on it being in ThirdParty is because otherwise you have to attach to each and every target - and this is not a pretty solution, it's ugly (not your fault, it's just one of the downfalls of CMake). I feel leaving it in the main CMakeLists.txt woud just clutter it and it would just be harder to maintain with it being in multiple places all over the project (we might have to update something or the other one day and we would have to do that 3 times).

beeperdeeper089: I'm willing to get this merged as is though, since my concern isn't blocking it's just something we deal with or don't, I would just recommend adding this part into the Client/Server CMakes too since they'll *definitely* need filesystem within the next couple PRs of work (saves us having to do it).

JosiahWI: I was going to wait to merge it in case I added more stuff, but I don't expect to do that in the near future now. I will look at this sometime this week, probably Saturday or Sunday, and add the flag to client and server. I also want to go back to your point about <filesystem> and MacOS. I will consider adding a check for that, although I can see why that's starting to get kinda messy if it's duplicated in three places. I could just put a compiler/OS check in the top level CMake and throw a fatal error with a more descriptive message than the linker error that would otherwise get thrown (not all gamers can read linker errors :-D).

beeperdeeper089: That sounds good. If you get around to it, hmu with a ping on discord since I haven't been super active on github for the last week and a bit or so (school started so been a bit busy :( )

beeperdeeper089 (approved):

beeperdeeper089:
I've approved this but I want to test this once on my own computer to make sure it works (CI doesn't pass in the first place so I would hope this works as intended).

JosiahWI:
Could we merge in the fixes to the sol2 submodule first, so the CI passes?

beeperdeeper089:
Well, we probably need CI on that PR to pass first.

beeperdeeper089:
@JosiahWI If you get develop re-merged into this, I can re-approve and get this merged since we have sol2 passing and we've switched to GH Actions for CI now.

JosiahWI:
Sounds great, I'll get it done today.

JosiahWI:
I was busy and didn't get this done yet, I apologise for my slothfulness on this PR and I won't let it happen again.

beeperdeeper089:
Don't worry about it, sometimes it happens. We're not really super concerned if it takes a few days for something to be done.

JosiahWI:
I made a simple mistake: I set the standard to C++11 and it should be C++17. I'll let you know when this is fixed.

Phoenix/Client/CMakeLists.txt
```diff
@@ -6,19 +6,21 @@ add_subdirectory(Source)
...
-set_target_properties(${PROJECT_NAME} PROPERTIES
-	CXX_STANDARD 17
-	CMAKE_CXX_STANDARD_REQUIRED ON
-	CMAKE_CXX_EXTENSIONS OFF
+	PRIVATE
+		Include
+		${PHX_THIRD_PARTY_INCLUDES}
 )
 
+target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_11)
```
beeperdeeper089, line 21: I'm pretty sure it's better for portability to use set_target_properties with CXX_STANDARD rather than that. Even if they're equivalent, just use set_target_properties, there's no particular reason to change what we have there.

Phoenix/Common/CMakeLists.txt
```diff
@@ -5,18 +5,22 @@ add_subdirectory(Source)
...
-	CMAKE_CXX_STANDARD_REQUIRED ON
-	CMAKE_CXX_EXTENSIONS OFF
+target_include_directories(${PROJECT_NAME}
+	PRIVATE
+		${PHX_THIRD_PARTY_INCLUDES}
+	PUBLIC
+		Include
 )
 
+target_compile_features(${PROJECT_NAME} PUBLIC cxx_std_11)
```
beeperdeeper089, line 21: Same as above.

Phoenix/ThirdParty/CMakeLists.txt
```diff
@@ -18,17 +15,22 @@ set(ALSOFT_TESTS OFF)
...
 set(BUILD_TESTING OFF)
 
-# define for EnTT to work
-add_definitions(-DNOMINMAX)
-
 add_subdirectory (openal-soft)
 add_subdirectory (SDL2)
+
+# Stop SDL trying to control our main function.
+target_compile_definitions(SDL2main INTERFACE SDL_MAIN_HANDLED)
```
beeperdeeper089, line 22: Maybe put this definition with the Entt one underneath, so they're in one place rather than in the middle of everything, resulting in some spaghetti.

beeperdeeper089 (changes requested):

beeperdeeper089:
Fix the cxx standard and these issues and we should be set to merge.

Phoenix/Client/CMakeLists.txt
```diff
@@ -6,17 +6,22 @@ add_subdirectory(Source)
...
+	PRIVATE
+		Include
+		${PHX_THIRD_PARTY_INCLUDES}
+)
 
 set_target_properties(${PROJECT_NAME} PROPERTIES
 	CXX_STANDARD 17
-	CMAKE_CXX_STANDARD_REQUIRED ON
-	CMAKE_CXX_EXTENSIONS OFF
+	CXX_STANDARD_REQUIRED ON
```
beeperdeeper089, line 23: Thanks for correcting that! Didn't even notice 😂

beeperdeeper089 (approved):

PR #287, opened 2020-09-15
Fix issue where empty folder is still recognised as a save by the Save system.

beeperdeeper089:
Resolves: #
Authors:
@beeperdeeper089 

## Summary of changes
- Checks for the existance of a json within the folder, rather than just the folder.
  
## Caveats
- It could be an issue if the folder doesn't have a json but has everything else. The save system might end up overwriting everything, however the likelyhood of this is low and I would be surprised if there is a valid situation for this to happen in.

## On approval
Merge.


beeperdeeper089:
Build will fail because Sol2, but it does work and it should work since it's just concatenating/appending the save name and ``.json``.

apachano (approved):

PR #302, opened 2020-10-28
Feat block models

beeperdeeper089:
Resolves: #
Authors:
@beeperdeeper089 

## Summary of changes
- Add block models
- Render block models
- Let block models be selected in lua.
  
## Caveats
- Still needs the new texturer to be implemented, but this can be merged first.

## On approval
Merge after Toby's approval after tests.


tobyplowy:
Can't add 5 textures to slope it only use first if u do. PLZ FEX!!ç!!!!!!!
(Yes this is a joke comment but it's still true)

Phoenix/Client/Include/Client/Graphics/BlockModel.hpp
```diff
@@ -0,0 +1,83 @@
...
+			return "Stair";
+		case BlockModel::X_PANEL:
+			return "XPanel";
+		case BlockModel::X_PANEL_CUBE:
+			return "XPanelCube";
+		}
+
+		// don't need this here but clang-tidy keeps complaining and we don't
+		// lose anything.
+		return "";
```
apachano, line 81: If we don't need a return value, why does the function have one?

apachano (approved):
Thanks for the giant PR ;)

Looks good tho, this functionality is great and I actually have some thoughts on how we can pull the vertices into a content file instead of compiled C if we want (Maybe . . . )

Issue #303, opened 2020-10-29
Slope block model can not have 5 textures in registration


tobyplowy:
## Describe the bug
A Slope has 5 sides. Each one should be able to have a texture. You can do this by adding 6 textures but it only has 5 sides so that isn't exactly right. It should be able to have 5 textures one for each side.

beeperdeeper089:
@tobyplowy Check out feat-textures on my fork and see if that seems to fix this issue.

PR #307, opened 2020-11-16
New and Improved texture handler.

beeperdeeper089:
Resolves: #303
Authors:
@beeperdeeper089 

## Summary of changes
- Replace old texturing system with a new modular one.
- Allowed multiple texture sizes to be done.
- Errors are handled a lot better - no random crashes if a texture isn't loaded right.
  
## Caveats
- Unknown textures haven't been setup yet (will do in a later PR once I have a better way to put the unknown texture data straight into code)
- Issue with texture wrapping with some seams in places. This is mostly fixed with ``GL_CLAMP_TO_EDGE`` as the wrapping option and anisotropic filtering. This can be further improved by anti-aliasing but I'll implement that better using a renderbuffer later on rather than relying on SDL's inbuilt stuff since we can then have some post-processing stuff too :).

## On approval
Merge, wait for @tobyplowy to give this a final test.

Testing Notes:
    - Make sure all textures are loading correctly in the right place in the right blocks.
    - Make sure the slope works, I remember fixing it but needs some more solid testing.
    - @tobyplowy you know what to do :wink:

### DIFF
--- a/Phoenix/Client/Include/Client/Graphics/CMakeLists.txt
+++ b/Phoenix/Client/Include/Client/Graphics/CMakeLists.txt
@@ -18,6 +18,7 @@ set(graphicsHeaders
 
 	${currentDir}/BlockModel.hpp
 	${currentDir}/BaseBlockModels.hpp
+	${currentDir}/TexturePacker.hpp
 	${currentDir}/ChunkMesher.hpp
 	${currentDir}/ChunkRenderer.hpp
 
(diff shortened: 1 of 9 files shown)


apachano (approved):
Looks good, runs good

tobyplowy:
Builds, But all textures are shuffled around. 
Grass reg has these textures. 
image
But it ends up looking like this. 
image
All blocks do this and yeah I did check the texture names to see if they are correct and they are both in the files and in lua.
My theory is that the game is stitching the textures at runtime and just not putting them in the right spot or it's chosing the wrong spots for the wrong blocks in the atlas. 
image
Furthermore... Stairs slopes and missing blocks still have the solid color texture issue. Tho for the slope it's only the bottom and slanted side.

tobyplowy (approved):
Looks good to me.
Builds fine and issues were resolved.

Issue #121, opened 2020-02-11
Metadata


apachano:
## Story
Sometimes we want custom data for a single block that differs from its BlockType, for this project we are dubbing that metadata. We need a system to store and interface with this metadata.

## MVP
What are the minimum features needed for this story?
- [ ] Blocks store metadata
- [ ] Metadata is accessible via getter and setter functions
- [ ] Metadata saves to file
- [ ] Blocks have no limit to how much metadata they can contain

## Stretch
What are some non-required features for this story?
- [ ] Lua can access metadata
- [ ] Items have similar metadata system

PR #315, opened 2020-11-29
Feat metadata

apachano:
Resolves: #121 
Authors:
@apachano 

## Summary of changes
- Implement a metadata storage system
- Integrate metadata storage system into blocks
- Record rotation on block placement in metadata on block
- Rotation is enabled during block-type registration
  
## Caveats
- Rotation does not render

## On approval
Merge


Phoenix/Common/Include/Common/Metadata.hpp
```diff
@@ -0,0 +1,80 @@
...
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+/**
```
beeperdeeper089, line 29: Do we want to bother standardising this? A few of our files have this but not that many.

Phoenix/Common/Include/Common/Metadata.hpp
```diff
@@ -0,0 +1,80 @@
...
+ * @brief Used to store unique data per individual instances
+ *
+ * @copyright Copyright (c) Genten Studios 2019 - 2020
+ *
+ */
+
+#pragma once
+
+#include <Common/Utility/Serializer.hpp>
+#include <any>
```
beeperdeeper089, line 40: Put a newline here just for consistency between our includes and standard ones (since that's what we're doing in most of our other code).

Phoenix/Common/Include/Common/Metadata.hpp
```diff
@@ -0,0 +1,80 @@
...
+#include <unordered_map>
+
+namespace phx
+{
+
+	class Metadata : public ISerializable
+	{
+	public:
+		/**
+		 * @brief sets or inserts metadata
```
beeperdeeper089, line 51: Being a bit pedantic (and I'm also a victim of this) but capitals and full stops.

Phoenix/Common/Include/Common/Metadata.hpp
```diff
@@ -0,0 +1,80 @@
...
+		 * @param key The key associated with the metadata
+		 * @return true if the data was set
+		 * @return false if the data already exists with a different data type
+		 */
+		bool set(const std::string& key, const std::any& existing);
+		/**
```
beeperdeeper089, line 57: Newline between functions.

Phoenix/Common/Include/Common/Metadata.hpp
```diff
@@ -0,0 +1,80 @@
...
+		 * @brief gets metadata by key
+		 * @param key The key associated with the metadata
+		 * @return A pointer to the data
+		 */
+		const std::any* get(const std::string& key) const;
+		/**
```
beeperdeeper089, line 63: Newline.

Phoenix/Common/Include/Common/Metadata.hpp
```diff
@@ -0,0 +1,80 @@
...
+		 * @brief Erases metadata
+		 * @param key The key associated with the metadata
+		 */
+		void erase(const std::string& key) { m_data.erase(key); };
+
+		size_t size() { return m_data.size(); };
```
beeperdeeper089, line 69: Use ``std::size_t``

Phoenix/Common/Include/Common/Metadata.hpp
```diff
@@ -0,0 +1,80 @@
...
+		 * @brief gets metadata by key
+		 * @param key The key associated with the metadata
+		 * @return A pointer to the data
+		 */
+		const std::any* get(const std::string& key) const;
+		/**
+		 * @brief Erases metadata
+		 * @param key The key associated with the metadata
+		 */
+		void erase(const std::string& key) { m_data.erase(key); };
```
beeperdeeper089, line 67: What happens if the key doesn't exist? (I think it just doesn't do anything, but would just like to double check)

Phoenix/Common/Source/Metadata.cpp
```diff
@@ -0,0 +1,125 @@
...
+Serializer& Metadata::operator>>(Serializer& ser) const
+{
+	int size = m_data.size();
+	ser << size;
+	for (auto& data : m_data)
+	{
+		ser << data.first;
+		if (data.second.type() == typeid(int))
+		{
+			ser << std::string("int") << std::any_cast<int>(data.second);
```
beeperdeeper089, line 68: 1. Do you have to construct as ``std::string`` - if so that is yet another shortcoming of the serializer -_-. If that needs fixing I can shove in a patch for that.

2. I would argue against putting in the full data type, just putting ``i, f, s, v (int, float, string, vector)`` would be a better idea.

Phoenix/Common/Source/Metadata.cpp
```diff
@@ -0,0 +1,125 @@
...
+			ser << std::string("float") << std::any_cast<float>(data.second);
+		}
+		else if (data.second.type() == typeid(phx::math::vec3))
+		{
+			math::vec3 v = std::any_cast<math::vec3>(data.second);
+			ser << std::string("vec3") << v.x << v.y << v.z;
+		}
+		else
+		{
+			LOG_FATAL("Metadata")
```
beeperdeeper089, line 81: We don't want to be able to get to a stage where this should ever happen - if there is an invalid type we've really fucked up somewhere but it shouldn't have to affect the general end user - so maybe this should be a debug message and it's just ignored otherwise?

Phoenix/Common/Include/Common/Voxels/Chunk.hpp
```diff
@@ -105,19 +106,38 @@ namespace phx::voxels
...
 
 		/**
-		 * @brief Sets the Block At the supplied position.
+		 * @brief Sets the Block at the supplied position.
 		 * @param position Position of the block relative to the chunk.
-		 * @param newBlock The block that exists at this location.
+		 * @param newBlock The block that now exists at this location.
 		 */
-		void setBlockAt(math::vec3 position, BlockType* newBlock);
+		void setBlockAt(math::vec3 position, Block newBlock);
```
beeperdeeper089, line 128: ``const math::vec3&``

Phoenix/Common/Include/Common/Voxels/Chunk.hpp
```diff
@@ -105,19 +106,38 @@ namespace phx::voxels
...
+		/**
+		 * @brief Sets metadata for the Block at the supplied position.
+		 * @param position Position of the block relative to the chunk.
+		 * @param key The key associated with the metadata.
+		 * @param newData The new value for the data.
+		 * @return true if the data was set.
+		 * @return false if the data already exists with a different data type
+		 * OR if the supplied position is out of bounds for the chunk.
+		 */
+		bool setMetadataAt(phx::math::vec3 position, const std::string& key,
```
beeperdeeper089, line 139: Const ref.

Phoenix/Common/Include/Common/Voxels/Map.hpp
```diff
@@ -77,7 +77,7 @@ namespace phx::voxels
 		static std::pair<math::vec3, math::vec3> getBlockPos(
 		    math::vec3 position);
 		BlockType* getBlockAt(math::vec3 position);
-		void       setBlockAt(math::vec3 pos, BlockType* block);
+		void       setBlockAt(math::vec3 pos, const Block& block);
```
beeperdeeper089, line 80: Const ref on vec3.

beeperdeeper089 (changes requested):
Some basic changes, however a concern at the base level:

  - If we're supporting limited types, like ints, floats, vec3 (which I would assume would make sense - bare in mind I haven't put much thought into this past now) - does std::any really make sense or should we use something like std::variant which supports limited types and makes it easier for us to track/label types within it?

Phoenix/Common/Include/Common/Metadata.hpp
apachano: I've been trying to, it does make the doxygen output nicer but its not super critical so I'm not concerned about pushing it if that makes sense?

Phoenix/Common/Source/Metadata.cpp
apachano: I totally spaced it in implementation but I feel we should keep fatal but kill the program. Leaving this unhanded will lead to corruption.

Phoenix/Common/Source/Metadata.cpp
apachano: You do have to construct as a string, if you don't then it will save as a cstring and in-turn will need retrieved as a cstring. Originally I put just a single character with the exception of a vec which may need expanded in the future to support additional types. 

We can convert this down to a single char - we just need to make a decision on how we would label a veci.

apachano:
@beeperdeeper089 I think I addressed everything, I also finished up everything I'm prepared to finish re rotation so there are a few changes I added making rotation optional.

Phoenix/Common/Source/Metadata.cpp
beeperdeeper089: Hmm, I'm not too sure, doing this leaves us in a spot of having many exit points through out the program (we definitely need to cut down on them), but I feel like metadata not being stored is not an issue enough that we should quit?

I'm not too sure, but my current argument is against exiting since I would assume it can't result in corruption or anything since it's just not outputting a piece of data.

beeperdeeper089 (commented):
Replied to a comment.

Phoenix/Common/Include/Common/Metadata.hpp
beeperdeeper089: Yeah sounds alright, I'll start telling Resharper to out that in my files too.

Phoenix/Common/Source/Metadata.cpp
apachano: Well its not just not outputting data, its not serializing a value that we will attempt to read later. If it doesn't segfault on read we will get a chunk full of OOB/ random blocks instead of what we would want. At the very least we should abort the save to avoid corruption but since this will never happen -> going any further than throwing an exception is a bit overkill.

Phoenix/Common/Source/Metadata.cpp
apachano: I converted to a single char, we can figure out handling multiple vec types later

Phoenix/Client/Include/Client/Voxels/BlockRegistry.hpp
```diff
@@ -129,16 +129,28 @@ namespace phx::client
...
+				    }
+
+				    sol::optional<bool> rotH = luaBlock["roth"];
+				    if (rotH)
+				    {
+					    block.rotH = *rotH;
+				    }
+
+				    sol::optional<bool> rotV = luaBlock["roth"];
+				    if (rotH)
```
beeperdeeper089, line 150: I'm assuming you meant ``if (rotV)``

Phoenix/Client/Include/Client/Voxels/BlockRegistry.hpp
```diff
@@ -129,16 +129,28 @@ namespace phx::client
...
+				    {
+					    block.rotH = *rotV;
```
beeperdeeper089, line 152: I'm assuming ``block.rotV``

beeperdeeper089 (changes requested):
You still have a few comments left to address and a few new ones.

apachano:
@beeperdeeper089 I think this is ready

tobyplowy:
Builds and runs. Can't find any issues.

beeperdeeper089:
Looks good, just still the use of ``std::size_t`` in Metadata.hpp:69.

I'll approve and merge once that's in.

apachano:
@beeperdeeper089 Fixed the size_t

beeperdeeper089 (approved):

PR #317, opened 2020-11-30
Feat hud

apachano:
Authors:
@apachano 

## Summary of changes
- Implements a super simple HUD that displays the block in the player's hand and the block the player is looking at

## Caveats
- No

## QA Notes
- HUD will display the block in the player's hand
- HUD will display the block the player is looking at, if there is no solid block in range the HUD will display air
- HUD will be located somewhere near the bottom-ish of the screen and will not be move-able
- Other UI interfaces won't be impacted

## On approval
Merge

### DIFF
--- a/Phoenix/Client/Include/Client/Game.hpp
+++ b/Phoenix/Client/Include/Client/Game.hpp
@@ -41,6 +41,7 @@
 #include <Client/InputQueue.hpp>
 #include <Client/Voxels/BlockRegistry.hpp>
 
+#include "HUD.hpp"
 #include <Common/CMS/ModManager.hpp>
 #include <Common/Save.hpp>
 
@@ -105,6 +106,7 @@ namespace phx::client
 
 		cms::ModManager* m_modManager;
 
+		HUD*        m_hud        = nullptr;
 		EscapeMenu* m_escapeMenu = nullptr;
 		GameTools*  m_gameDebug  = nullptr;
 		bool        m_followCam  = true;
--- a/Phoenix/Client/Include/Client/Game.hpp
+++ b/Phoenix/Client/Include/Client/Game.hpp
@@ -38,10 +38,10 @@
 #include <Client/Graphics/ShaderPipeline.hpp>
 #include <Client/Graphics/Window.hpp>
 #include <Client/Graphics/ChatBox.hpp>
-#include <Client/InputQueue.hpp>
 #include <Client/Voxels/BlockRegistry.hpp>
+#include <Client/InputQueue.hpp>
+#include <Client/HUD.hpp>
 
-#include "HUD.hpp"
 #include <Common/CMS/ModManager.hpp>
 #include <Common/Save.hpp>
 
--- a/Phoenix/Client/Source/HUD.cpp
+++ b/Phoenix/Client/Source/HUD.cpp
@@ -1,0 +1,67 @@
+// Copyright 2019-20 Genten Studios
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice,
+// this list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice,
+// this list of conditions and the following disclaimer in the documentation
+// and/or other materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its contributors
+// may be used to endorse or promote products derived from this software without
+// specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+#include <Client/Client.hpp>
+#include <Client/HUD.hpp>
+
+#include <Common/Actor.hpp>
+#include <Common/PlayerView.hpp>
+
+#include <imgui.h>
+
+using namespace phx::client;
+using namespace phx;
+
+HUD::HUD(gfx::Window* window, entt::registry* registry, entt::entity player)
+    : gfx::Overlay("HUD"), m_window(window), m_registry(registry),
+      m_player(player)
+{
+}
+
+void HUD::onAttach() {}
+void HUD::onDetach() {}
+void HUD::onEvent(events::Event& e) {}
+
+void HUD::tick(float dt)
+{
+	voxels::Map* map = m_registry->get<PlayerView>(m_player).map;
+	ImGui::SetNextWindowPos(
+	    {m_window->getSize().x / 2 - 250, m_window->getSize().y - 150},
+	    ImGuiCond_Once);
+
+	ImGui::Begin("HUD", nullptr, ImGuiWindowFlags_NoResize);
+	{
+		ImGui::Text("%s",
+		            map->getBlockAt(ActorSystem::getTarget(m_registry, m_player)
+		                                .getCurrentPosition())
+		                ->displayName.c_str());
+		ImGui::Text("Hand: %s",
+		            m_registry->get<Hand>(m_player).hand->displayName.c_str());
+	}
+	ImGui::End();
+}
--- a/Phoenix/Client/Source/HUD.cpp
+++ b/Phoenix/Client/Source/HUD.cpp
@@ -51,15 +51,24 @@ void HUD::tick(float dt)
 {
 	voxels::Map* map = m_registry->get<PlayerView>(m_player).map;
 	ImGui::SetNextWindowPos(
-	    {m_window->getSize().x / 2 - 250, m_window->getSize().y - 150},
+	    {m_window->getSize().x / 2 - 250, m_window->getSize().y - 100},
 	    ImGuiCond_Once);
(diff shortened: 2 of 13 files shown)


Phoenix/Client/Include/Client/Game.hpp
```diff
@@ -41,6 +41,7 @@
 #include <Client/InputQueue.hpp>
 #include <Client/Voxels/BlockRegistry.hpp>
 
+#include "HUD.hpp"
```
beeperdeeper089, line 44: Use Client/Include/...

beeperdeeper089 (changes requested):

Phoenix/Client/Include/Client/Game.hpp
apachano: Fixed . . . I must've let the IDE dot hat one :/

Phoenix/Client/Source/HUD.cpp
```diff
@@ -0,0 +1,67 @@
...
+	ImGui::SetNextWindowPos(
+	    {m_window->getSize().x / 2 - 250, m_window->getSize().y - 150},
+	    ImGuiCond_Once);
+
+	ImGui::Begin("HUD", nullptr, ImGuiWindowFlags_NoResize);
+	{
+		auto target =
+		    ActorSystem::getTarget(m_registry, m_player).getCurrentPosition();
+		target.floor();
+		ImGui::Text("%s", map->getBlockAt(target)->displayName.c_str());
```
beeperdeeper089, line 62: What happens in the case of an out-of-bounds or air block? Does it display that too? (Sorry, haven't had the chance to test this out yet.)

Phoenix/Client/Source/HUD.cpp
```diff
@@ -0,0 +1,67 @@
...
+void HUD::onEvent(events::Event& e) {}
+
+void HUD::tick(float dt)
+{
+	voxels::Map* map = m_registry->get<PlayerView>(m_player).map;
+	ImGui::SetNextWindowPos(
+	    {m_window->getSize().x / 2 - 250, m_window->getSize().y - 150},
+	    ImGuiCond_Once);
+
+	ImGui::Begin("HUD", nullptr, ImGuiWindowFlags_NoResize);
```
beeperdeeper089, line 57: Like Toby said on Discord, do we want to add whatever the flag is for disabling movement of the window?

beeperdeeper089 (commented):
Everything looks fine, I've just got two questions :)

Phoenix/Client/Source/HUD.cpp
beeperdeeper089: Also, do we want to set an explicit size for the ImGui window? Maybe it can be either a set size or relative to the window size (problem with it being relative is if someone decides to scale the window down it might completely mess it up)

Gerold55: I think it should be disabled by default but there should be a way to be able to activate it for people making mods imo...

Phoenix/Client/Source/HUD.cpp
apachano: It should display an OOB (those are solids right?), it will display an air block if there is no solid within range of the player.

beeperdeeper089: Do we want it to - as a user facing HUD, would it make sense to hide things like OOB and air?

Phoenix/Client/Source/HUD.cpp
apachano: We probably want to set an explicit size once we know how wide it needs to be, although there isn't anything in-particular wrong with it setting the size based on container contents.

I can hunt down the flag and add it, not critical IMO but probably something we do want along with removing the title and possibly making the background transparent. Lots of style stuff that will eventually need to be added.

As far as modding support we could build that in later, thats definitely going to be a deeper discussion figuring out how we want to disable movement in those scenarios (which will likely be done via lua and probably repositioning instead of free movement?)

beeperdeeper089: Sounds good to me 👍

Phoenix/Client/Source/HUD.cpp
apachano: You can break them so they should be solids, this should probably be the implementation anyways as the end user will want to break OOBs.

Air on the other hand - probably not.

beeperdeeper089: Why would the user want to break OOB's? An OOB should only ever return on an error that is that the block you're trying to get doesn't actually exist - so why should they be able to break it? That would end up breaking a lot of systems and design if we want that to be a thing. I would assume we don't want things like Out of Bounds being usable and just an error type. I might be understanding wrong but this is what I've come to understand of an OOB block.

apachano: I mean - if you can't break it what do you do with it? Is it just a piece of the chunk you can never use again? An OOB is just to highlight that there was a block there but now it doesn't exist (usually because a module was removed or a definition was changed). Alternatively I'd argue that we would display an OOB as a proxy to air so the end-user never sees it and we never save it.

beeperdeeper089: Ah, in the case you're highlighting, if a block no longer exists, due to something like a mod being removed, it would've replaced by core.unknown. Out of Bounds should only be returned if you're requesting a block that is out of the bounds of the map or a specific chunk. (this block doesn't necessarily have to exist, however it is a preventative measure in the case that it occurs.)

apachano: OHHHHH got it lol, well in that case it can't highlight an OOB and we are all good. I just got the two mixed up.

apachano:
Okay @beeperdeeper089 I locked + centered the window and added logic to not display non-solid blocks.

Phoenix/Client/Source/HUD.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+void HUD::onAttach() {}
+void HUD::onDetach() {}
+void HUD::onEvent(events::Event& e) {}
+
+void HUD::tick(float dt)
+{
+	voxels::Map* map = m_registry->get<PlayerView>(m_player).map;
+	ImGui::SetNextWindowPos(
+	    {m_window->getSize().x / 2 - 250, m_window->getSize().y - 100},
```
beeperdeeper089, line 54: these are magic numbers, just make like a ``static constexpr int HUD_WIDTH`` variable or something so it makes sense when we come back to it.

Phoenix/Client/Source/HUD.cpp
```diff
@@ -0,0 +1,76 @@
...
+	    ImGuiCond_Once);
+	ImGui::SetNextWindowSize(ImVec2(500, 50), ImGuiCond_Once);
```
beeperdeeper089, line 56: Same as above, but you might need to put HUD_WIDTH / 2 for above.

Phoenix/Client/Source/HUD.cpp
```diff
@@ -0,0 +1,76 @@
...
+		std::string target_display;
+		auto        target =
+		    ActorSystem::getTarget(m_registry, m_player).getCurrentPosition();
+		target.floor();
+		auto target_block = map->getBlockAt(target);
+		if (target_block->category == voxels::BlockCategory::SOLID)
+		{
+			target_display = target_block->displayName;
+		}
+		ImGui::Text("%s", target_display.c_str());
```
beeperdeeper089, line 70: Can we not just do:
```
if (target_block->category == voxels::BlockCategory::SOLID)
{
    ImGui::Text(..., targetBlock->displayName)
}
```
This prevents you copying the string unnecessarily.

Phoenix/Client/Source/HUD.cpp
```diff
@@ -0,0 +1,76 @@
...
+	ImGui::SetNextWindowSize(ImVec2(500, 50), ImGuiCond_Once);
+
+	ImGui::Begin("HUD", nullptr,
+	             ImGuiWindowFlags_NoDecoration | ImGuiWindowFlags_NoMove);
+	{
+		std::string target_display;
+		auto        target =
+		    ActorSystem::getTarget(m_registry, m_player).getCurrentPosition();
+		target.floor();
+		auto target_block = map->getBlockAt(target);
```
beeperdeeper089, line 65: Style, ``targetBlock``

Phoenix/Client/Source/HUD.cpp
```diff
@@ -0,0 +1,76 @@
...
+	voxels::Map* map = m_registry->get<PlayerView>(m_player).map;
+	ImGui::SetNextWindowPos(
+	    {m_window->getSize().x / 2 - 250, m_window->getSize().y - 100},
+	    ImGuiCond_Once);
+	ImGui::SetNextWindowSize(ImVec2(500, 50), ImGuiCond_Once);
+
+	ImGui::Begin("HUD", nullptr,
+	             ImGuiWindowFlags_NoDecoration | ImGuiWindowFlags_NoMove);
+	{
+		std::string target_display;
```
beeperdeeper089, line 61: I've recommended a way for you to get rid of this below, but if you keep this, needs to be renamed to ``targetDisplay``

beeperdeeper089 (changes requested):

Phoenix/Client/Source/HUD.cpp
apachano: It couldn't be static since its calculated but I can create an extra variable for it

Phoenix/Client/Source/HUD.cpp
apachano: @beeperdeeper089 Then the UI will re-adjust every time you are/aren't looking at something solid because instead of filling it with an empty text element we don't add anything and everything below it slides up.

Phoenix/Client/Source/HUD.cpp
JosiahWI: Not the function call part of the express though unless it's a constexpr function.

Phoenix/Client/Source/HUD.cpp
apachano: Oh I see what you are saying looking at this tonight

apachano:
@beeperdeeper089 I think this is ready

tobyplowy:
@apachano Builds and runs. 
Possible issue. Not really a big issue as this is temp and not meant to be final but still pointing it out for the record...
Changing the window size changes the position of the hotbar. (Tested on windows)
image
image

Phoenix/Client/Source/HUD.cpp
```diff
@@ -0,0 +1,78 @@
...
+void HUD::onAttach() {}
+void HUD::onDetach() {}
+void HUD::onEvent(events::Event& e) {}
+
+void HUD::tick(float dt)
+{
+	voxels::Map* map = m_registry->get<PlayerView>(m_player).map;
+	ImGui::SetNextWindowPos(
+	    {m_window->getSize().x / 2 - WIDTH / 2, m_window->getSize().y - POSY},
+	    ImGuiCond_Once);
```
beeperdeeper089, line 55: Removing ImGuiCond_Once would fix Toby's issue. Also, will using it here prevent any future SetNextWindowPos calls? From what I read of the documentation it would, but it doesn't highlight whether it actually does or not so we should probably check or just remove it since I would assume the performance difference is negligible (we're not using ImGui particularly extensively in the first place).

beeperdeeper089 (commented):
One comment, but otherwise this is ready to merge.

Phoenix/Client/Source/HUD.cpp
beeperdeeper089: ^ Just like ``static constexpr int HUD_WIDTH = 500``

Phoenix/Client/Source/HUD.cpp
apachano: It doesn't prevent any, just that particular call won't call again. I forgot why I added it though.

beeperdeeper089 (approved):
Lgtm

PR #321, opened 2020-12-20
Feat items

apachano:
Authors:
@apachano 

## Summary of changes
- Implements base framework for item implementation
- Implements item registration through Lua
- Switches player hand from blocks to items
- Implements stored value for blocks dropping items
  
## Caveats
- Item and block places/drops store strings as arguments to allow for registration without the target existing
- Block drops functionality is not actually implemented

## QA Notes
- Functionally nothing should be different than before
- Only registered items are available for the hand selector, blocks should not be

## On approval
Merge

### DIFF
--- a/Phoenix/Common/Include/Common/Voxels/Item.hpp
+++ b/Phoenix/Common/Include/Common/Voxels/Item.hpp
@@ -61,8 +61,8 @@ namespace phx::voxels
 	{
 	public:
 		// the base ItemReferrer object implements this.
-		static constexpr std::size_t UNKNOWN_ITEM       = 0;
-		static constexpr std::size_t OUT_OF_BOUNDS_ITEM = 1;
+		static constexpr std::size_t UNKNOWN_ITEM = 0;
+		static constexpr std::size_t INVALID_ITEM = 1;
 
 	public:
 		ItemType()  = default;
--- a/Phoenix/Common/Include/Common/Voxels/Item.hpp
+++ b/Phoenix/Common/Include/Common/Voxels/Item.hpp
@@ -1,0 +1,107 @@
+// Copyright 2019-20 Genten Studios
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice,
+// this list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice,
+// this list of conditions and the following disclaimer in the documentation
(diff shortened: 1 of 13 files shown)


tobyplowy:
LGTM. Builds, runs, items work. Good job!

Phoenix/Common/Include/Common/Voxels/Item.hpp
```diff
@@ -0,0 +1,107 @@
...
+	 * "Type" regardless of the number of instances of that item in the
+	 * world. For example a dirt block will have the same basic information
+	 * every time it exists in the world.
+	 */
+	struct ItemType
+	{
+	public:
+		// the base ItemReferrer object implements this.
+		static constexpr std::size_t UNKNOWN_ITEM       = 0;
+		static constexpr std::size_t OUT_OF_BOUNDS_ITEM = 1;
```
beeperdeeper089, line 65: Only skimmed over this, but what's the OUT_OF_BOUNDS_ITEM for?

Phoenix/Common/Source/Actor.cpp
```diff
@@ -163,34 +163,44 @@ bool ActorSystem::action2(entt::registry* registry, entt::entity entity)
...
 			{
-				math::vec3 rotation = {};
-				if (dir.x > 0)
+				voxels::Block block {m_blockReferrer->getByID(item->places),
+				                     nullptr};
+				Metadata      data;
+				if (block.type->rotH)
 				{
-					rotation.x += 180;
+					math::vec3 rotation = {};
```
beeperdeeper089, line 174: You don't need to initialize this, it default constructs.

Phoenix/Common/Include/Common/Voxels/Item.hpp
apachano: Inventories, if you request an item in a slot the inventory doesn't have you get this.

Phoenix/Common/Source/Actor.cpp
apachano: Fixed

Phoenix/Common/Include/Common/Voxels/Item.hpp
beeperdeeper089: Maybe it's better to call it an "Invalid Item" instead? Because there aren't really any bounds so imo the naming doesn't make much sense.

apachano: Sounds good to me - updated

beeperdeeper089 (approved):

PR #331, opened 2021-01-14
New Settings Systems with support for multiple data types.

beeperdeeper089:
Resolves: #
Authors:
@beeperdeeper089 

## Summary of changes
- Allow the settings system to use booleans, strings, and any arithmetic type.
- Exit on failure when there is an invalid config file.
- Use the config file provided in the CLI arguments.

## Caveats
- No more settings in the settings exit menu (will attempt reimplement before merging)
- No more settings registration in the Lua modules (will attempt reimplemntation before merging)

## On approval
Wait for me.


Phoenix/Client/Source/EscapeMenu.cpp
```diff
@@ -91,19 +93,20 @@ void EscapeMenu::tick(float dt)
...
+		//for (const auto& setting : settings)
+		//{
+		//	int i = setting.second.value();
+		//	ImGui::SliderInt(setting.second.getName().c_str(), &i,
+		//	                 setting.second.getMin(), setting.second.getMax());
+		//	if (i != setting.second.value())
+		//	{
+		//		Settings::get()->getSetting(setting.second.getKey())->set(i);
+		//	}
+		//}
```
apachano, line 108: We will want to patch this up before merging

Phoenix/Client/Source/Game.cpp
```diff
@@ -80,7 +80,6 @@ Game::Game(gfx::Window* window, entt::registry* registry, bool networked)
 		m_network->sendMessage(text);
 	});
 
-	Settings::get()->registerAPI(m_modManager);
```
apachano, line 84: We want to finish this portion as well

Phoenix/Client/Source/Graphics/Camera.cpp
```diff
@@ -35,7 +35,9 @@ using namespace phx;
 using namespace gfx;
 
 FPSCamera::FPSCamera(Window* window, entt::registry* registry)
-    : m_registry(registry), m_enabled(true), m_window(window)
+    : m_window(window), m_enabled(true), m_settingSensitivity(
+	      Settings::instance()->getOr("camera:sensitivity", 50.f)),
```
apachano, line 39: So we are defining the 50.f in two places now - How can we ensure that both get updated at the same time if we ever decide to change them, what's the source of truth here?

Phoenix/Common/Include/Common/Settings.hpp
```diff
@@ -36,166 +36,298 @@
...
+		 * first and then get it - if the validation returns false however,
+		 * skipVerif must not be enabled otherwise there will be issues
+		 * during usage.
 		 *
-		 * @param value The minimum value the setting can be.
 		 */
-		void setMin(int value);
-		int  getMin() const { return m_minValue; }
+		template <typename Type>
+		Setting<Type> get(const std::string& key, bool skipVerif = false)
```
apachano, line 163: Why is this defined inline?

apachano (commented):
Handful of comments, mostly the settings file shouldn't be defined inline and we need to re-attach everything that was commented out so we don't loose a ton of functionality. 

Since we are lacking tests, it would also be beneficial to demo the setting functionality for all supported types in the lua files.

Phoenix/Common/Include/Common/Settings.hpp
beeperdeeper089: Because it's a templated function.

Phoenix/Client/Source/Graphics/Camera.cpp
beeperdeeper089: Removed the other one, that should never have an issue since this should be initialized before something like the escape menu.

apachano (commented):
The code is okay but you need to make mac happy

Some thought re lua: first Lua itself only supports one numeral type, a double, so we should probably consider scrapping float and int support from the modding interface as lua will just convert it anyways.

We should also think thorough a getorset() method for lua or overloading the function so we can get without supplying a default value. In general we should be encouraging setting defaults in one location and not forcing them to be set in multiple (potentially out of sync) locations.

apachano (dismissed):

beeperdeeper089:
Had I known Lua only supports doubles....

Regardless, the get method can work just fine without a default parameter - if the value doesn't exist it will default initialize (so most likely, to 0), however getOr implements what you're saying. I may not have that exposed properly to the Lua interface - but I'll take a look and see if I can drop that in easily.

Mac is very... retarded - it's giving me an issue with std::size_t being an unsigned long instead of an unsigned long long -_-

tobyplowy:
Build failed. 
[message (3).txt](https://github.com/GentenStudios/Phoenix/files/5838206/message.3.txt)

apachano (dismissed):
LGTM

beeperdeeper089:
ci can fuck itself

apachano (dismissed):

tobyplowy:
Still fails to build for me. 
[message (4).txt](https://github.com/GentenStudios/Phoenix/files/5862621/message.4.txt)

tobyplowy:
Builds but has that crash on startup thing.

tobyplowy:
Builds and runs. Congrats it was quite the journey.

apachano (approved):

Issue #194, opened 2020-04-23
Improve saving and loading systems.


tobyplowy:
## Story
Right now the saving system is a bit... Questionable, but it does the job for now tho we for sure want to upgrade it sometime soonish. 

This may be useful. 
https://www.seedofandromeda.com/blogs/1-creating-a-region-file-system-for-a-voxel-game

## MVP
-  [ ] Feature 1

apachano:
A couple of additional thoughts

- We don't need to store consecutive and redundant blocks. If a chunk is all dirt, we can save that as `core::dirt^4096;` or something similar instead of listing each block.
- We will need to store metadata in 0.3
- We **MUST** store the uniqueName and not an integer as all integers are invalidated between runtimes if the modules change.

apachano:
I marked the checkbox on binary storage since it was added when metadata was introduced.

PR #348, opened 2021-01-31
Serialize consecutive metadata-less blocks with a quantity specifier.

apachano:
Resolves: None - Checks box 1 on #194 
Authors:
@apachano 

## Summary of changes
- When serializing a chunk, if two or more blocks have the same ID and no metadata we serialize a '*' character followed by the number of blocks that meet this criteria massively reducing the data used for chunks of mostly the same block. 
  
## Caveats
None

## QA Notes
- Save file size should be reduced
- Everything else should function the same

## On approval
Merge

### DIFF
--- a/Phoenix/Common/Source/Voxels/Chunk.cpp
+++ b/Phoenix/Common/Source/Voxels/Chunk.cpp
@@ -102,7 +102,8 @@ phx::Serializer& Chunk::operator>>(phx::Serializer& ser) const
 		{
 			std::size_t j;
 			for (j = 1; i + 1 < CHUNK_MAX_BLOCKS &&
-			            m_blocks[i + 1]->id == m_blocks[i]->id;
+			            m_blocks[i + 1]->id == m_blocks[i]->id &&
+			            m_metadata.find(i + 1) == m_metadata.end();
 			     j++)
 			{
 				i++;
--- a/Phoenix/Common/Source/Voxels/Chunk.cpp
+++ b/Phoenix/Common/Source/Voxels/Chunk.cpp
@@ -103,6 +103,7 @@ phx::Serializer& Chunk::operator>>(phx::Serializer& ser) const
 		         m_metadata.find(i + 1) == m_metadata.end())
 		{
 			std::size_t j = i;
+			i++;
 			while (i + 1 < CHUNK_MAX_BLOCKS &&
 			       m_blocks[i + 1]->id == m_blocks[i]->id &&
 			       m_metadata.find(i + 1) == m_metadata.end())
--- a/Phoenix/Common/Source/Voxels/Chunk.cpp
+++ b/Phoenix/Common/Source/Voxels/Chunk.cpp
@@ -113,7 +113,6 @@ phx::Serializer& Chunk::operator>>(phx::Serializer& ser) const
 		else if (canRepeat(i))
 		{
 			std::size_t j = i;
-			i++;
 			while (canRepeat(i))
 			{
 				i++;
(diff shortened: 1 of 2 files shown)


Phoenix/Common/Source/Voxels/Chunk.cpp
```diff
@@ -98,6 +98,18 @@ phx::Serializer& Chunk::operator>>(phx::Serializer& ser) const
 		{
 			ser << '+' << m_metadata.at(i);
 		}
+		else if (m_blocks[i + 1]->id == m_blocks[i]->id)
```
JosiahWI, line 101: This is inside a for loop that iterates from 0 to the last array index. Isn't index i + 1 going to run off the end of the sequence if the last block has no metadata?

Phoenix/Common/Source/Voxels/Chunk.cpp
```diff
@@ -98,6 +98,18 @@ phx::Serializer& Chunk::operator>>(phx::Serializer& ser) const
...
+		{
+			std::size_t j;
+			for (j = 1; i + 1 < CHUNK_MAX_BLOCKS &&
```
JosiahWI, line 104: Suggestion here; you could set j to i and increment j until you find metadata, then serialize the difference between j and i and set i back to j. It would save incrementing i each iteration.

Phoenix/Common/Source/Voxels/Chunk.cpp
```diff
@@ -129,6 +141,16 @@ phx::Serializer& Chunk::operator<<(phx::Serializer& ser)
 			ser >> data;
 			m_metadata[i] = data;
 		}
+		else if (c == '*')
+		{
+			std::size_t rep;
+			ser >> rep;
+			for (std::size_t j = 1; j < rep; j++)
```
JosiahWI, line 148: If rep == 2, this will only iterate once, but shouldn't it iterate twice?

Phoenix/Common/Source/Voxels/Chunk.cpp
JosiahWI: Setting j to i + 2*

Phoenix/Common/Source/Voxels/Chunk.cpp
apachano: No, because we already stored the first block in the sequence in the core segment of the de-serialization on line 131 so we are starting with line 2 explicitly.

Phoenix/Common/Source/Voxels/Chunk.cpp
apachano: Yea it will :( in all the test-cases the second to last block was always within a sequence so I never caught it, I'll add an extra check here.

Phoenix/Common/Source/Voxels/Chunk.cpp
apachano: Good Idea! I switched to a while loop and keep incrementing i instead of j + setting i in the end but the logic is the same.

JosiahWI (dismissed):
I haven't tested this or checked it carefully against the coding style, but it looks good to me.

apachano:
@tobyplowy I can't repro your issue, can you try again after the latest changes to see if you can? (Also start with a fresh save on this one since it breaks save compatibility to an extent)

tobyplowy:
Still same thing even with a fesh save.
image
Here's my save in case of nay use. [save.zip](https://github.com/GentenStudios/Phoenix/files/5901239/save.zip)

tobyplowy:
Maybe it's a winderps specific issue?

apachano:
Probably Windows specific - but I think I know where its doing that and might be able to fix it

tobyplowy:
Nope still same thing.

apachano:
I hate your computer

tobyplowy:
That did the trick.

beeperdeeper089 (approved):
I *think* this looks fine, i guess it'll come up somewhat soon in the future if its not perfect.

PR #349, opened 2021-02-03
Fix missing include for iomanip in Settings.cpp

JosiahWI:
Resolves: Compilation error in GCC 3.8 from undefined setw symbol.
Authors:
@JosiahWI 

## Summary of changes
- The iomanip header is included in Settings.cpp
  
## Caveats
- None

## On approval
Merge


tobyplowy:
Builds and runs not like it wouldn't but still worth checking none the less.

beeperdeeper089 (approved):

PR #351, opened 2021-02-03
Use the array delete instead of the improper delete

JosiahWI:
Resolves: When deleting an array delete[] should be used or it will not be properly deleted.
Authors:
@JosiahWI

## Summary of changes
- An array allocated by the ChatBox is now deleted using delete[] instead of delete.
  
## On approval
Merge


apachano (approved):
LGTM

PR #355, opened 2021-02-14
Group third party libraries together in an interface target

JosiahWI:
Resolves: 
Authors:
@JosiahWI 

## Summary of changes
- The variables for third party libraries and headers were removed, and replaced with a single
- interface target PhoenixThirdParty that links the appropriate libraries.
- Some include directories are also added to the target because their respective projects were
- too lazy to add include directories to their targets.
  
## On approval
Merge


tobyplowy:
Builds and runs as intended.

beeperdeeper089 (approved):
Beautiful, ty for this @JosiahWI!

PR #359, opened 2021-02-26
Fix segfault on re-opening escape menu

beeperdeeper089:
Resolves: #
Authors:
@beeperdeeper089 

## Summary of changes
- Doesn't remove the EscapeMenu from the stack anymore, but rather just disables it and re-enables it when necessary.
  
## Caveats
- None.

## On approval
Merge.

### DIFF
--- a/Phoenix/Client/Include/Client/EscapeMenu.hpp
+++ b/Phoenix/Client/Include/Client/EscapeMenu.hpp
@@ -57,6 +57,8 @@ namespace phx::client
 		void tick(float dt) override;
 
 		bool isActive() { return m_active; };
+		void enableMenu();
+		void disableMenu();
 
 		static constexpr float WIDTH  = 300;
 		static constexpr float HEIGHT = 300;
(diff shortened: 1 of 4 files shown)


tobyplowy:
Builds and issue seems to be fixed.

apachano:
@beeperdeeper089 So why don't we want to remove the escape menu from the layer stack? This seems like it defeats the purpose of having the layer stack in the first place.

beeperdeeper089:
Yeah, use cases like this is what I overlooked during the initial revision of this. The layer stack takes control of the memory, so as soon as you push a layer to the layer stack, the stack can call delete the layer as it wants when the conditions are right, so it ended up deleting the escape menu, what we can do is remove that idea of deleting and just have the layer stack pop/push layers and work the top one.

Due to this, one of my next PRs implements a GameStateManager which is just a layer stack but without the complexity of overlays and memory management. The only thing with it is since overlays aren't a thing, each individual state has to render it's overlays (e.g. Game renders Crosshair and Escape menu as it needs within the game class rather than pushing an overlay, as annoying as that sounds it makes it a lot easier to handle the states of each).

apachano (approved):
Gotcha, we will chat in discord further then and can just pull this in as a hotfix. Its definitely not the ideal endstate however.

PR #364, opened 2021-05-18
Refactor Skybox and Selectionbox renderers into their own objects.

apachano:
Authors: @apachano 

## Summary of changes
- Refactors WorldRenderer.hpp into SkyboxRenderer.hpp with the intent that the object only handles the skybox.
- Splits SelectionboxRenderer.hpp from ChunkRenderer.hpp with the intent that the chunk renderer only renders chunks.
- Removes Camera from ChunkRenderer.hpp logic.
  
## Caveats
- The chunk renderer needs further cleanup, its definitely still messy.
- A pointer to the map is stored in both the Position and PlayerView, however PlayerView likely needs depreciated as switching between entities will cause problems regardless. 

## Testing notes
Nothing functionally should have changed, this is purely a code reorganization in preparation for adding an additional renderer.

## On approval
Merge

### DIFF
--- a/Phoenix/Client/Include/Client/Graphics/SelectionboxRenderer.hpp
+++ b/Phoenix/Client/Include/Client/Graphics/SelectionboxRenderer.hpp
@@ -32,6 +32,7 @@
 
 #include <Common/Position.hpp>
 
+#include <entt/entity/registry.hpp>
 #include <string>
 #include <vector>
 
@@ -47,7 +48,7 @@ namespace phx::gfx
 		 * @param projection Calculated projection matrix.
 		 * @param selection Position of block to be highlighted.
 		 */
-        void tick(const Position& position, const math::mat4& projection, math::vec3 selection);
+        void tick(entt::registry* registry, entt::entity entity, const math::mat4& projection);
 	private:
         unsigned int   m_selectionBoxVAO = 0;
         unsigned int   m_selectionBoxVBO = 0;
--- a/Phoenix/Client/Include/Client/Graphics/SelectionboxRenderer.hpp
+++ b/Phoenix/Client/Include/Client/Graphics/SelectionboxRenderer.hpp
@@ -1,0 +1,56 @@
+// Copyright 2019-20 Genten Studios
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice,
+// this list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice,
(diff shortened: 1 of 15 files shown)


tobyplowy:
PR builds and runs.

Phoenix/Client/Include/Client/Graphics/SelectionboxRenderer.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include <Common/Position.hpp>
+
+#include <entt/entity/registry.hpp>
+#include <string>
+#include <vector>
+
+namespace phx::gfx
+{
+	class SelectionboxRenderer
```
beeperdeeper089, line 41: ``SelectionBoxRenderer`` for PascalCase consistency.

Phoenix/Client/Include/Client/Graphics/SelectionboxRenderer.hpp
```diff
@@ -0,0 +1,57 @@
...
+	public:
+		SelectionboxRenderer();
+		/**
+		 * @brief Highlights a block for the rendered frame.
+		 * @param position Position of the entity viewing the block.
+		 * @param projection Calculated projection matrix.
+		 * @param selection Position of block to be highlighted.
+		 */
+        void tick(entt::registry* registry, entt::entity entity, const math::mat4& projection);
+	private:
```
beeperdeeper089, line 52: Newline for styling + my ocd

beeperdeeper089 (changes requested):

Phoenix/Common/Include/Common/Actor.hpp
```diff
@@ -85,7 +84,12 @@ namespace phx
 
 		static math::Ray    getTarget(entt::registry* registry,
 		                              entt::entity    entity);
-		static entt::entity registerActor(entt::registry* registry);
+
+        static entt::entity registerActor(
+		    entt::registry* registry,
+		    voxels::Map*    map,
+		    math::vec3      position = {0,0,0},
```
beeperdeeper089, line 91: Are we sure we want to allow default initialisation of the position and rotation?
I feel like that should always be explicitly set just due to the fact that 0, 0, 0 could spawn you inside the world randomly and could give us a bug we didn't forsee later.

beeperdeeper089 (changes requested):

Phoenix/Common/Include/Common/Actor.hpp
tobyplowy: Yeah Vyom has a point here.

apachano: That makes sense, I'll remove the default initialisation

beeperdeeper089 (approved):
LGTM.

PR #368, opened 2021-06-18
Consolidate duplicate code into single function

apachano:
Authors:
@apachano 

## Summary of changes
- Consolidate duplicate code into single function
  
## Caveats
- A constructor making GL calls may be an issue

## On approval
Merge

### DIFF
--- a/Phoenix/Client/Include/Client/Graphics/ChunkRenderer.hpp
+++ b/Phoenix/Client/Include/Client/Graphics/ChunkRenderer.hpp
@@ -57,6 +57,22 @@ namespace phx::gfx
 		math::vec3 normal;
 	};
 
+    /**
+     * @brief A struct to store the data required to render chunks.
+     *
+     * Don't worry about this if you don't have OpenGL knowledge, this is
+     * mainly an internal data format anyway.
+     */
+    struct ChunkRenderData
+    {
+        /// @brief The vertex array object.
+        unsigned int vao;
+        /// @brief The buffer object.
+        unsigned int vbo;
+        /// @brief The amount of vertices to render.
+        std::size_t vertexCount;
+    };
+
 	/**
 	 * @brief Renders submitted chunks, and allows for dropping and updating of
 	 * chunks.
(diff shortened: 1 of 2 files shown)


Phoenix/Client/Include/Client/Graphics/ChunkRenderer.hpp
```diff
@@ -155,6 +156,6 @@ namespace phx::gfx
 		static const int m_normalAttributeLocation = 2;
 		static const int m_colorAttributeLocation  = 3;
 
-        gfx::ShaderPipeline m_renderPipeline;
+        gfx::ShaderPipeline m_renderPipeline{};
```
beeperdeeper089, line 159: Do we actually need to default init this? The class does have a default constructor, and do we want to use ``= {}`` cos imo it does look tidier?

Phoenix/Client/Include/Client/Graphics/ChunkRenderer.hpp
```diff
@@ -104,6 +88,24 @@ namespace phx::gfx
 	class ChunkRenderer : public voxels::MapEventSubscriber
 	{
 	public:
+        /**
+         * @brief A struct to store the data required to render chunks.
+         *
+         * Don't worry about this if you don't have OpenGL knowledge, this is
+         * mainly an internal data format anyway.
+         */
+        struct ChunkRenderData
```
beeperdeeper089, line 97: This is fine, but delete the copy constructor and copy and move assignment operators so we don't get weird resource lifetime bugs one day due to copies and deletes

Phoenix/Client/Include/Client/Graphics/ChunkRenderer.hpp
```diff
@@ -104,6 +88,24 @@ namespace phx::gfx
...
+        {
+			ChunkRenderData() = default;
+            explicit ChunkRenderData(std::vector<float> mesh);
+            /// @brief The vertex array object.
+            unsigned int vao{};
```
beeperdeeper089, line 102: And do we want to use ``= {}`` cos imo it looks tider.

beeperdeeper089 (changes requested):

beeperdeeper089:
Like you said, constructor making GL calls could get annoying, so just mark the coup constructor as delete as well as the move and copy assignment operators, you can leave the move constructor as default since that shouldn't interfere.

Phoenix/Client/Include/Client/Graphics/ChunkRenderer.hpp
apachano: idk, clang-tidy was throwing a fit that it wasn't ever initialized so I rolled with it

apachano:
@beeperdeeper089 Alternatively it may make sense for this to be a private member function on the renderer itself? That was my original plan but for some reason I liked this last night

beeperdeeper089:
I think it may make sense as a private member, in my opinion RAII for OpenGL is quite a non-intuitive way of working it bc it's just horrid for it 😂

Phoenix/Client/Include/Client/Graphics/ChunkRenderer.hpp
```diff
@@ -113,10 +114,13 @@ namespace phx::gfx
 
 		void prep();
 
+    private:
+        static ChunkRenderData generate(std::vector<float> mesh);
```
beeperdeeper089, line 118: ``const ref`` or r-value ref to move, otherwise we're copying a potentially somewhat large amount of data

beeperdeeper089 (changes requested):

beeperdeeper089 (approved):

PR #369, opened 2021-06-19
Replace JSON submodule with static version

beeperdeeper089:
Resolves: #
Authors:
@beeperdeeper089 

## Summary of changes
- JSON submodule is massive and can take over a minute to clone, so just use static files since it's not getting updated super often anyway.
  
## Caveats
- None, updating is a slight bit more work but their releases should be a simple unzip and drag and drop.

## On approval
Merge.


apachano (approved):
Approved, but does this really save download time? Its the same files either way right so wouldn't this just increase the download time for our repo by a minute?

beeperdeeper089:
The git history for the JSON submodule makes it massive, this is just a mere megabyte or two.

